PR #5, opened 2021-05-27
Another Round of Cleanup

hexonaut:
Small things:

 * DC adjustments for test as Aave market size has grown
 * Inline interfaces to `FooLike` style to match MIPs guidelines
 * bool -> uint256 to save gas
 * Allow setting the target for stkAAVE in case we move it to a management contract later
 * Dont allow `cull()` on global settlement
 * Allow DAI holders to `exit()` during GS


KenG-pro (approved):

derek-flossman:
Added recommended changes for Global Settlement, will require a review time-permitting.

hexonaut:
Going to merge this to prepare for V1 of the review.

PR #9, opened 2021-09-14
Team Review Changes

hexonaut:


hexonaut:
Note for reviewers: Due to rebasing nature of token the deposit/withdraw, interactions with adai may be off by 1 unit. In particular:

[`pool.deposit`](https://github.com/makerdao/dss-direct-deposit/pull/9/files#diff-c07f3e830c4fa78152c08182fbf4cca1366646ca164433876cec57b69a521b10R234) will round against the user due to this line: https://etherscan.io/address/0x7b2a3cf972c3193f26cdec6217d27379b6417bd0#code#F1#L139

The scaled balance is divided by the interest index which is then later re-multiplied to give back the actual balance. Error should be at most `1` against our contract.

gbalabasquer:
I know that unwinding DAI would be almost immediately but in case that can't happen for any reason I think we should cover the case of ES.
So being strictly specific we couldn't keep unwinding until `end.debt > 0`. But that would mean that we need to add the `end` dependency to this module. So I'd suggest we could stop unwinding if `vat.live == 0`.

gbalabasquer:
Well maybe the benefit of getting rid of `aDai` in case of ES makes it worth it to check the `End.debt` value.

gbalabasquer:
- Having this function: `function unwind(uint256 amount, uint256 fees) external {` we don't really need to leave the path of using `exec` when `live == 0`. Or otherwise we could just make this one `auth`.

gbalabasquer:
- Do we really need `emergencyExit` having `quit`? It's true that `emergencyExit` has a `wad` amt but is it really useful?
I like to cover all the emergency exit paths but I think we should try to leave what is just necessary.

gbalabasquer:
Yeah actually discussing it with Kurt, he pointed out that unwinding after ES actually won't work as it is due the `frob` but also because anyone can skim the `urn` taking out all the `aDai` from this module.
So we need to better evaluate ES.
Good about the other points, but let's discuss them deeper on Tuesday's call.

PR #10, opened 2021-09-28
Changes for ES + other minor things

gbalabasquer:


src/DssDirectDepositAaveDai.sol
```diff
@@ -381,20 +467,21 @@ contract DssDirectDepositAaveDai {
...
         require(live == 0, "DssDirectDepositAaveDai/live");
-        require(_add(tic, tau) <= block.timestamp, "DssDirectDepositAaveDai/early-cull");
+        require(_add(tic, tau) <= block.timestamp || wards[msg.sender] == 1, "DssDirectDepositAaveDai/early-cull");
         require(culled == 0, "DssDirectDepositAaveDai/already-culled");
 
         (uint256 ink, uint256 art) = vat.urns(ilk, address(this));
         require(ink <= 2 ** 255, "DssDirectDepositAaveDai/overflow");
         require(art <= 2 ** 255, "DssDirectDepositAaveDai/overflow");
-        vat.grab(ilk, address(this), address(this), address(chainlog.getAddress("MCD_VOW")), -int256(ink), -int256(art));
+        vat.grab(ilk, address(this), address(this), chainlog.getAddress("MCD_VOW"), -int256(ink), -int256(art));
```
brianmcmichael, line 476: I would almost rather the `vow` be fileable than rely on the chainlog for calls like this.

Using the chainlog here requires that it becomes a test dependency, it just seems cleaner to use a storage slot for this.

src/DssDirectDepositAaveDai.sol
```diff
@@ -381,20 +467,21 @@ contract DssDirectDepositAaveDai {
...
+
         culled = 1;
         emit Cull();
     }
 
     // --- Emergency Quit Everything ---
     function quit(address who) external auth {
-        require(live == 0, "DssDirectDepositAaveDai/live");
+        require(vat.live() == 1, "DssDirectDepositAaveDai/no-quit-during-shutdown");
```
brianmcmichael, line 484: `live` is changed to `0` in `cull()` which would be required for this to be called before. Just want to clarify the intent to change the function here.

src/DssDirectDepositAaveDai.sol
iamchrissmith: for my learning: did you remove `address` because it would already be returned as one and no need to cast?

brianmcmichael: Yeah the return is already an address. See: https://github.com/makerdao/dss-direct-deposit/pull/10/files#R30

src/DssDirectDepositAaveDai.sol
```diff
@@ -381,20 +467,21 @@ contract DssDirectDepositAaveDai {
     function cull() external {
         require(vat.live() == 1, "DssDirectDepositAaveDai/no-cull-during-shutdown");
         require(live == 0, "DssDirectDepositAaveDai/live");
-        require(_add(tic, tau) <= block.timestamp, "DssDirectDepositAaveDai/early-cull");
+        require(_add(tic, tau) <= block.timestamp || wards[msg.sender] == 1, "DssDirectDepositAaveDai/early-cull");
```
iamchrissmith, line 470: nit: change message to `"DssDirectDepositAaveDai/unauthorized-cull"`

iamchrissmith (changes requested):
I have a few minor comments here.  My main thing is really struggling with the introduction of so many logic branches based on the mode.  But I want to see if I can parse/organize them locally before I take a stab at a comment on them here

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
+        return _rdiv(_add(stableDebt.totalSupply(), variableDebt.totalSupply()), targetUtil);
     }
+
+    // --- Deposit controls ---
     function _wind(uint256 amount) internal {
+        if (amount == 0) {
+            return;
+        }
+
+        bytes32 ilk_ = ilk;
```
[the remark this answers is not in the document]
brianmcmichael: This gets used a few times later. By doing this we only do a single SLOAD into a memory variable which can be used repeatedly later at less expense.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
 
-        _wind(amount);
+        // Do inverse calculation of interestStrategy
+        uint256 targetUtil;
+        if (targetInterestRate > _add(interestStrategy.baseVariableBorrowRate(), interestStrategy.variableRateSlope1())) {
+            // Excess interest rate
+            uint256 r = targetInterestRate - interestStrategy.baseVariableBorrowRate() - interestStrategy.variableRateSlope1();
+            targetUtil = _add(_rdiv(_rmul(interestStrategy.EXCESS_UTILIZATION_RATE(), r), interestStrategy.variableRateSlope2()), interestStrategy.OPTIMAL_UTILIZATION_RATE());
+        } else {
+            // Optimal interst rate
```
kmbarry1, line 248: "interst" --> "interest"

src/DssDirectDepositAaveDai.sol
```diff
@@ -246,94 +282,144 @@ contract DssDirectDepositAaveDai {
...
+            daiDebt = vat.gem(ilk_, address(end));
+        } else if (mode == 4) {
+            // MCD caged but previously module was caged and culled
+            // debt is obtained from free collateral owned by the End module after some processing
+            daiDebt = vat.gem(ilk_, address(this));
+            end = chainlog.getAddress("MCD_END");
+            if (daiDebt > 0) {
+                require(daiDebt < 2 ** 255, "DssDirectDepositAaveDai/overflow");
+                vat.grab(ilk_, address(this), address(this), vow, int256(daiDebt), int256(daiDebt));
+                EndLike(end).skim(ilk_, address(this));
```
kmbarry1, line 314: I _think_ we could just `flux` the `gem` to the End (saves some gas)...unless there's some possibility that `tag` for this `ilk` will not be one. Doesn't hurt to do it the proper way, I suppose.

src/DssDirectDepositAaveDai.sol
```diff
@@ -246,94 +282,144 @@ contract DssDirectDepositAaveDai {
...
-            vat.frob(ilk, address(this), address(this), address(this), -int256(amount), -int256(amount));
+
+        if (mode == 1) {
+            vat.frob(ilk_, address(this), address(this), address(this), -int256(amount), -int256(amount));
+            vat.slip(ilk_, address(this), -int256(amount));
+            vat.move(address(this), vow, _mul(fees, RAY));
+        } else if (mode == 2) {
+            vat.slip(ilk_, address(this), -int256(amount));
+            vat.move(address(this), vow, _mul(total, RAY));
+        } else if (mode == 3 || mode == 4) {
```
kmbarry1, line 365: nit: could express condition more simply as `mode > 2`

src/DssDirectDepositAaveDai.sol
```diff
@@ -202,10 +207,13 @@ contract DssDirectDepositAaveDai {
...
         z = _mul(x, RAY) / y;
     }
+    function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = x <= y ? x : y;
+    }
 
     // --- Administration ---
     function file(bytes32 what, uint256 data) external auth {
-        require(live == 1, "DssDirectDepositAaveDai/not-live");
+        require(vat.live() == 1, "DssDirectDepositAaveDai/no-file-during-shutdown");
```
hexonaut, line 216: This check should not replace the original `live` check. It should be in addition.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
 
         emit File(what, data);
     }
+
     function file(bytes32 what, address data) external auth {
+        require(vat.live() == 1, "DssDirectDepositAaveDai/no-file-during-shutdown");
```
hexonaut, line 229: We don't need to block changing this during GS. This is for Aave rewards and has nothing to do with GS.

src/DssDirectDepositAaveDai.sol
gbalabasquer: or just `else` :)

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yes there is an easier way to do it with `slip`, but as you said doing via the proper methods looks more correct.
Regarding a different `tag`, this whole module assumes two things are true:
1) `rate` will be always 1
2) price, then `tag` will be always 1 (or maybe "at least 1")

For the second one, if it is lower than 1 we will start filling in `End.gap[ilk]`, which will provoke that we can't `slip` from the End anymore for unwinding as we will fuck up the whole accounting. So I think we better keep those two premises always true for this module. We should document them well.

src/DssDirectDepositAaveDai.sol
hexonaut: We've discussed this before. I really think it's a mistake to have to synchronously update the `vow` in every module if/when we update it. Having a dynamic lookup ensures we don't miss anything on the periphery.

src/DssDirectDepositAaveDai.sol
hexonaut: I think this might have been a mistake on the original PR. You should be able to quit at any time during non-ES operation.

hexonaut (changes requested):
Some requested changes.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
+            targetUtil = _rdiv(_rmul(_sub(targetInterestRate, interestStrategy.baseVariableBorrowRate()), interestStrategy.OPTIMAL_UTILIZATION_RATE()), interestStrategy.variableRateSlope1());
+        }
+        return _rdiv(_add(stableDebt.totalSupply(), variableDebt.totalSupply()), targetUtil);
     }
+
+    // --- Deposit controls ---
     function _wind(uint256 amount) internal {
+        if (amount == 0) {
+            return;
+        }
```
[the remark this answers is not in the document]
hexonaut: Aave `pool.deposit()` will error if the amount to deposit is 0. This check needs to be moved below line 266 anyways. It's causing the `test_hit_debt_ceiling` test to fail.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yeah I'm still dubious about it as for example in `dss-deploy-scripts` I had a problem with relying the changelog to deploy contracts. The changelog will be something that we will probably update after redeploying (like we did in Goerli). As if the modules requires the chainlog to load addresses we are in a problem for that case.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
+
         if (what == "king") king = data;
         else revert("DssDirectDepositAaveDai/file-unrecognized-param");
         emit File(what, data);
     }
 
-    // --- Deposit controls ---
-    function wind(uint256 amount) external auth {
-        require(live == 1, "DssDirectDepositAaveDai/not-live");
+    // --- Automated Rate Targetting ---
```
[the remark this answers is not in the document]
gbalabasquer: Fixed.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Fixed.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Fixed.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Hmm what is the intention to block that?
I mean if the module is caged, and governance decides to change the `bar` nothing happens as still goes through the other paths.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Well we need to assume a malicious governance. We do not want they have access to change the `king`, however is quite probably the `king` will be the pause proxy, so they will still get the rewards.
Anyway I think is more about a good practice to block all the authed functions if the `vat` is caged.
But not feeling too strong about this specific case either (and we haven't followed that pattern in other contracts either).

src/DssDirectDepositAaveDai.sol
gbalabasquer: Changed to `else`

src/DssDirectDepositAaveDai.sol
```diff
@@ -246,94 +282,144 @@ contract DssDirectDepositAaveDai {
...
-    function unwind(uint256 amount, uint256 fees) external {
-        require(wards[msg.sender] == 1 || live == 0, "DssDirectDepositAaveDai/not-authorized");
+    function _unwind(uint256 supplyReduction, uint256 availableLiquidity, uint256 mode) internal {
+        bytes32 ilk_ = ilk;
+        address vow = chainlog.getAddress("MCD_VOW");
+        address end;
+        uint256 adaiBalance = adai.balanceOf(address(this));
+
+        uint256 daiDebt;
+        if (mode == 1) {
```
talbaneth, line 292: Suggestion - would be nice to use enums for these modes.
For example instead of mode "1"/"2"/.. we will have something like:
Mode.ONLY_CAGED
Mode.CULLED
Mode.MCD_CAGED
Mode.CULLED_THEN_MCD_CAGED

src/DssDirectDepositAaveDai.sol
gbalabasquer: We will keep discussing this as I think the usage of the chainlog for the `end` is even more controversial.
But that can be addressed in a different PR.

src/DssDirectDepositAaveDai.sol
hexonaut: Right, fair point about attack. This is a pretty edge case that in practise will likely not amount to a large amount of funds since they are likely already sent and this thing will probably be closed out mostly during ES. I'm fine if you want to leave as-is.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yeah that's true, we could actually remove the check at all.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Modified to use enum.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
+        bytes32 ilk_ = ilk;
+
+        // Wind amount is limited by the debt ceiling
+        (uint256 Art,,, uint256 line,) = vat.ilks(ilk_);
+        uint256 lineWad = line / RAY; // Round down to always be under the actual limit
+        uint256 newDebt = _add(Art, amount);
+        if (newDebt > lineWad) amount = _sub(lineWad, Art);
+
+        if (amount == 0) {
+            return;
```
julienmartinlevrai, line 265:
```suggestion
            emit Wind(amount);
            return;
```

src/DssDirectDepositAaveDai.sol
```diff
@@ -246,94 +282,143 @@ contract DssDirectDepositAaveDai {
...
+            fees = adaiBalance - daiDebt;
+
+            if (_add(amount, fees) > availableLiquidity) {
+                // Don't need safe-math because this is constrained above
+                fees = availableLiquidity - amount;
+            }
+        }
+
+        if (amount == 0 && fees == 0) {
+            return;
```
julienmartinlevrai, line 347:
```suggestion
            emit Unwind(amount);
            return;
```

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +224,53 @@ contract DssDirectDepositAaveDai {
...
         emit File(what, data);
     }
 
-    // --- Deposit controls ---
-    function wind(uint256 amount) external auth {
-        require(live == 1, "DssDirectDepositAaveDai/not-live");
+    // --- Automated Rate targeting ---
+    function calculateTargetSupply(uint256 targetInterestRate) public view returns (uint256) {
+        require(targetInterestRate > 0, "DssDirectDepositAaveDai/target-interest-zero");
+        require(targetInterestRate <= interestStrategy.getMaxVariableBorrowRate(), "DssDirectDepositAaveDai/above-max-interest");
```
julienmartinlevrai, line 239: These two checks are redundant when the function is called from `exec`. Are they intended for external calls ?

src/DssDirectDepositAaveDai.sol
```diff
@@ -103,6 +103,11 @@ interface RewardsClaimerLike {
     function getRewardsBalance(address[] calldata assets, address user) external view returns (uint256);
```
julienmartinlevrai, line 103:
```suggestion
```

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +225,53 @@ contract DssDirectDepositAaveDai {
...
     }
+
+    // --- Deposit controls ---
     function _wind(uint256 amount) internal {
+        bytes32 ilk_ = ilk;
+
+        // Wind amount is limited by the debt ceiling
+        (uint256 Art,,, uint256 line,) = vat.ilks(ilk_);
+        uint256 lineWad = line / RAY; // Round down to always be under the actual limit
+        uint256 newDebt = _add(Art, amount);
```
julienmartinlevrai, line 262: In order to help future maintainers avoid hard-to-see bugs, it is important to make our assumptions explicit.

Here, we are assuming `rate = 1`, therefore normalized debt equals debt (ignoring units), therefore this addition makes sense.

Note also that computing a WAD line is uncommon in our codebase. Why is that ? Because usually we compare it to `art * rate`, which is a RAD. So, in a way, when we divide line by `RAY`, we are actually dividing it by `rate`.

My main concern is that new développers look at this and don’t see anything unusual, and then miss the distinction between normalized debt and régular debt where it matters.

I would add a comment explaining this assumption. Or, better still, I would refactor the code so that debt is measured in `RAD`, as is the practice in our core modules.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +225,53 @@ contract DssDirectDepositAaveDai {
...
+        if (newDebt > lineWad) amount = _sub(lineWad, Art);
+
```
julienmartinlevrai, line 264:
```suggestion

        if (newDebt > availableLiquidity) {
           amount = _sub(availableLiquidity, Art); 
        }

```

julienmartinlevrai (commented):
I have been thinking about the best path forward in a bank run scénario.

It is tricky, because we want to provide liquidity précisely when there is not enough, but at a certain point we have to stop providing it and start trying to recover it.

Currently, in a bank run scénario, we will provide liquidity until we hit `line`. There will be no way to prévent that, and if Aave doesn’t recover from it, we will only be able to recover what is left after an emergency spell is cast.

In order to prévent that, we should:

a) be able to permissonlessly `cage` the module if its debt is higher than the available liquidity, thus being able to `unwind` at once whatever we can get.
b) not `wind` more than we can `unwind`.

The most important considération against adding these checks, is if they are likely to result in false positives during times of high demand for DAI.

I added code suggestions for both checks — please regard them as pseudocode. If consensus is reached for adding them, I can work on a PR with extra tests for this scénario.

src/DssDirectDepositAaveDai.sol
gbalabasquer: The check was removed from the file that sets the `bar`.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Hmm do we want to log this no-op? hmmm maybe yes but not 100% sure.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Same here.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Hmmm the `>0` one you are right, but the other is not totally the case. `interestStrategy.getMaxVariableBorrowRate()` could change after you set `bar`.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Actually this will break as it is being used in the test. But I agree we shouldn't leave interfaces in the main code for the test. Will decouple them.

src/DssDirectDepositAaveDai.sol
julienmartinlevrai: I just checked [the code](https://etherscan.io/address/0xfffE32106A68aA3eD39CcCE673B646423EEaB62a#code), the only way for it to change is if the `interestStategy` changes, which would allow permissionlessly `cage`ing the module.

src/DssDirectDepositAaveDai.sol
julienmartinlevrai: It will make life easier for people writing scripts, less corner cases to consider as the function will either revert or emit.

src/DssDirectDepositAaveDai.sol
gbalabasquer: I've added some comments, let me know what you think and if we should expand better and in which places.

julienmartinlevrai: I like them, thank you !

iamchrissmith:
> a) be able to permissonlessly `cage` the module if its debt is higher than the available liquidity, thus being able to `unwind` at once whatever we can get. 
> b) not `wind` more than we can `unwind`.

Overall, I like these suggestions; however, I would say `a` should not be `cage` since that is a one-way street.  Could we instead have `exec` auto-unwind some amount when `daiDebt > availableLiquidity`?

julienmartinlevrai:
That is an interesting suggestion @iamchrissmith, and it actually touches on `b)` as well, since in some cases the `wind` amount should not only be smaller, but negative — in other words, the module should unwind, which makes me think the change should happen in `exec`.

gbalabasquer:
@iamchrissmith @hexonaut @talbaneth @julienmartinlevrai @brianmcmichael @kmbarry1 I think the PR is probably in a good stage to start making a deep review. I might still miss some minor thing or add some other test, but should be fine for you to look at it.
Btw I removed one of the modes and made a new function `uncull`. I think it will be better to have a simple function to move the collateral back to the position just in case there isn't availability for unwind. This way we make sure we have the collateral ready to be skimmed and go to the end module to be shared in case something is going wrong with the exec function.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yeah that's right, still not sure if we should keep it just in case a future strategy does it a bit differently. I'll let @hexonaut to decide on this one.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Same here, @hexonaut what do you think?

src/DssDirectDepositAaveDai.t.sol
```diff
@@ -505,7 +561,260 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
+
+        // We try to unwind what is possible
+        deposit.exec();
+        VowAbstract(vow).heal(min(vat.sin(vow), vat.dai(vow)));
+
+        // A part can't be unwind yet
+        assertEq(vat.gem(ilk, address(end)), amountSupplied / 2);
+        assertGt(adai.balanceOf(address(deposit)), amountSupplied / 2);
+        assertLe(vat.sin(vow), originalSin + part * RAY - originalDai - (amountSupplied / 2) * RAY);
+        assertGe(vat.sin(vow), originalSin + part * RAY - originalDai - (amountSupplied / 2 + 1) * RAY);
```
julienmartinlevrai, line 779: ├╴call 0x35D1b3F3D7966A1DFe207aa4514C12a259A0492B::sin(@0xA950524441892A31ebddF91d3cEEFa04Bf454466) (src/DssDirectDepositAaveDai.t.sol:779)
      │  └╴← (0)
      ├╴log("Error: a >= b not satisfied [uint]") (lib/ds-test/src/test.sol:224)
      ├╴log_named_uint("  Value a", 0) (lib/ds-test/src/test.sol:225)
      ├╴log_named_uint("  Value b", 115792089237316195423570961774909335961466340321992856468664914230636615358125) (lib/ds-test/src/test.sol:226)

src/DssDirectDepositAaveDai.sol
hexonaut: I don't think we want to restrict ourselves with this. It's quite likely the total vault debt will be higher than the available liquidity during normal operation, and it doesn't really prevent Aave users from removing the available Dai liquidity later.

src/DssDirectDepositAaveDai.sol
```diff
@@ -216,25 +209,59 @@ contract DssDirectDepositAaveDai {
...
+            targetUtil = _add(_rdiv(_rmul(interestStrategy.EXCESS_UTILIZATION_RATE(), r), interestStrategy.variableRateSlope2()), interestStrategy.OPTIMAL_UTILIZATION_RATE());
+        } else {
+            // Optimal interest rate
+            targetUtil = _rdiv(_rmul(_sub(targetInterestRate, interestStrategy.baseVariableBorrowRate()), interestStrategy.OPTIMAL_UTILIZATION_RATE()), interestStrategy.variableRateSlope1());
+        }
+        return _rdiv(_add(stableDebt.totalSupply(), variableDebt.totalSupply()), targetUtil);
     }
+
+    // --- Deposit controls ---
     function _wind(uint256 amount) internal {
```
hexonaut, line 240: Nit: These functions don't need leading underscores anymore.

src/DssDirectDepositAaveDai.sol
```diff
@@ -246,94 +273,139 @@ contract DssDirectDepositAaveDai {
...
+            (daiDebt,) = vat.urns(ilk_, address(this));
+        } else if (mode == Mode.MODULE_CULLED) {
+            // Module shutdown and culled
+            // debt is obtained from free collateral owned by this contract
+            daiDebt = vat.gem(ilk_, address(this));
+        } else {
+            // MCD caged
+            // debt is obtained from free collateral owned by the End module
+            end = chainlog.getAddress("MCD_END");
+            EndLike(end).skim(ilk_, address(this));
```
hexonaut, line 297: Wanting to verify `skim()` can be called multiple times. This appears to be the case, but just double checking.

hexonaut (changes requested):
Partial review. Will finish tomorrow. `test_unwind_culled_then_mcd_caged` seems to be failing as well.

src/DssDirectDepositAaveDai.sol
julienmartinlevrai: I like it because they are internal

src/DssDirectDepositAaveDai.sol
gbalabasquer: Done.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Done.

src/DssDirectDepositAaveDai.sol
```diff
@@ -202,11 +194,12 @@ contract DssDirectDepositAaveDai {
     function _rdiv(uint256 x, uint256 y) internal pure returns (uint256 z) {
         z = _mul(x, RAY) / y;
     }
+    function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
```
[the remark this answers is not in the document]
gbalabasquer: Done.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yeah let's keep it as with the rest of internals.

src/DssDirectDepositAaveDai.sol
gbalabasquer: Yeah, `skim()` can be called several times, but after the CDP is grabbed, it will be a no op.
For a normal CDP this will happen the first time, in this case it would also happen the first time if the deposit contract isn't culled. In that case the `skim` that will count will be the first one after `uncull` is called.

src/DssDirectDepositAaveDai.t.sol
```diff
@@ -505,7 +561,260 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
+        // Then pool gets some liquidity
+        pool.repay(address(dai), amountToBorrow, 2, address(this));
+
+        // Rest of the liquidity can be withdrawn
+        deposit.exec();
+        VowAbstract(vow).heal(min(vat.sin(vow), vat.dai(vow)));
+        assertEq(vat.gem(ilk, address(end)), 0);
+        assertEq(adai.balanceOf(address(deposit)), 0);
+        assertEq(vat.sin(vow), 0);
+        assertGe(vat.dai(vow), originalDai - originalSin + daiEarned * RAY);
```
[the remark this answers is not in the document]
gbalabasquer: Yeah I've been having this 1 RAY rounding issue on different places. Which is coming from 1 wei rounding converted to RAD.
In this case I'm checking:
```
assertGe(vat.dai(vow), originalDai - originalSin + daiEarned * RAY);
assertLe(vat.dai(vow), originalDai - originalSin + (daiEarned + 1) * RAY);
```
Wondering if should be instead: 
```
assertGe(vat.dai(vow), originalDai - originalSin + (daiEarned - 1) * RAY);
assertLe(vat.dai(vow), originalDai - originalSin + daiEarned * RAY);
```
But I think I had the case where I got +1 error. Maybe should be between -1 and +1.

src/DssDirectDepositAaveDai.sol
brianmcmichael: Yeah it's a lot easier to identify internal calls elsewhere in the code this way.

src/DssDirectDepositAaveDai.t.sol
```diff
@@ -505,7 +561,260 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
+        // CDP still has the position built
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(deposit));
+        assertGt(ink, 0);
+        assertGt(art, 0);
+        assertEq(vat.gem(ilk, address(end)), 0);
+
+        // We try to unwind what is possible
+        deposit.exec();
+
+        // Part can't be done yet
```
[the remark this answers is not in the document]
gbalabasquer: Fixed the comment and added the `vat.sin(vow)` check.

src/DssDirectDepositAaveDai.t.sol
```diff
@@ -505,7 +561,260 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
+        assertEq(art, 0);
+        assertEq(vat.gem(ilk, address(end)), amountSupplied / 2); // Automatically skimmed when unwinding
+
+        // Some time later the pool gets some liquidity
+        hevm.warp(block.timestamp + 180 days);
+        pool.repay(address(dai), amountToBorrow, 2, address(this));
+
+        // Rest of the liquidity can be withdrawn
+        deposit.exec();
+        assertEq(vat.gem(ilk, address(end)), 0);
```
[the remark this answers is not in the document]
gbalabasquer: Done.

src/DssDirectDepositAaveDai.sol
```diff
@@ -363,12 +450,12 @@ contract DssDirectDepositAaveDai {
...
     function cage() external {
-        // Can shut this down if we are authed, if the vat was caged
+        require(vat.live() == 1, "DssDirectDepositAaveDai/no-cage-during-shutdown");
+        // Can shut this down if we are authed
         // or if the interest rate strategy changes
         (,,,,,,,,,, address strategy,) = pool.getReserveData(address(dai));
         require(
             wards[msg.sender] == 1 ||
-            vat.live() == 0 ||
             strategy != address(interestStrategy)
```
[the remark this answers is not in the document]
gbalabasquer: Marking this solved, as if happens will be in a separate PR.

src/DssDirectDepositAaveDai.t.sol
gbalabasquer: I think this should be fixed now using directly the `assertEqApprox` function.

src/DssDirectDepositAaveDai.t.sol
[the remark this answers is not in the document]
gbalabasquer: This should be fixed now.

PR #11, opened 2021-10-05
Prevent a bank run

julienmartinlevrai:
Prévent a banque run by never allowing the `targetSupply` to surpass `availableLiquidity`.

This is ready for review, except for the current tests, which are failing, and the bank run tests, which I haven’t done yet.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,25 @@
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository and submodules
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Install nix 2.3.6
+      uses: cachix/install-nix-action@v13
+      with:
+        install_url: https://releases.nixos.org/nix/nix-2.3.6/install
+        nix_path: nixpkgs=channel:nixos-unstable
+
+    - name: Use maker and dapp cachix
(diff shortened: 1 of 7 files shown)


hexonaut:
I'm not quite sure what the goal of this PR is. The way it is currently written it will ensure the module is almost never used since `availableLiquidity` is almost always around 20% of the total `supplyAmount` and subsequently `targetSupply`. This will almost certainly always result in attempts to unwind the position no matter the current level. Are you wanting to limit the total debt of the vault to be under the `availableLiquidity`? In that case that is a different equation. It would be something like:

```
(uint256 daiDebt,) = vat.urns(ilk, address(this));
int256 delta = int256(targetSupply) - int256(supplyAmount);
uint256 newDebt = uint256(int256(daiDebt) + delta);
if (newDebt > availableLiquidity) {
   // Do re-calc of targetSupply or whatever
}
```

If this is what you were going for (which I'm assuming based on our RC convo), then I would prefer not to add this restriction. It will limit our exposure to be about 20% of the current total debt of the pool and possibly run into some unintuitive behaviours such as when there is a surge in demand we cannot fill it as quickly. We would have to call `exec()` multiple times in this scenario to marginally increase the `availableLiquidity` when we could have just called it once. We can limit total exposure via the debt ceiling same as any other collateral.

I suppose in the event of a catastrophic bug in Aave the Dai where all the Dai is removed from the pool the D3M would stop minting more, but this could be circumvented by flash loan, mint adai, trigger exec, burn adai, repay flash loan to hit the debt ceiling anyways.

PR #12, opened 2021-10-06
Add pie parameter as brakes on bank run exposure

iamchrissmith:
## Problem
In the case of a bank run, we would expect the D3M's share of the deposited Dai to increase.  

## Approach
`pie` allows governance to set a maximum share of the pool that it is comfortable exposing Maker to through the D3M.

## Details
This is different than the `line` since the `line` is an absolute value.  This is relative to the other participants in the pool.  In the case of a bank run without `pie` we would expect to see the following scenario:

- Depositors withdraw as much Dai as they can (eat up all the liquidity)
- Depositors are then incentivized to call `exec` on the D3M to `wind` up to `line`
- This creates more liquidity in the pool so depositors can take more Dai out and leave the unclaimable `aDai` in the D3Ms balance.

Since `line` is an absolute value the bigger D3M and the Aave Dai pool are, the more risk is there is that D3M and Maker will be left holding the bag in the case of a bank run.

By adding `pie` we can enforce not just the absolute value of `line` for our `aDai` exposure, but also the relative exposure.

In the above scenario as the other depositors withdraw their Dai, they will be leaving the D3M with a great share of the overall pie.  Eventually even though the `line` may not have been met the `pie` will be reached and our exposure in the bank run will be limited.

Early on, it could be that this is simply 100% (1 `RAY`) since our `line` will be low and governance might be comfortable having full exposure to the `line` in the case of a bank run, but if the system grows, governance may decide to set a very high `line` to allow for the D3M to float with the "normal" markets and `wind`/`unwind` as needed even to large numbers, but in the case of a bank run, it would provide an braking mechanism as the system unwinds giving governance time to cage this, or set high `bar`s so we `unwind` (or manually call `unwind`)

### DIFF
--- a/src/DssDirectDepositAaveDai.sol
+++ b/src/DssDirectDepositAaveDai.sol
@@ -125,6 +125,7 @@ contract DssDirectDepositAaveDai {
 
     uint256 public immutable tau;   // Time until you can write off the debt [sec]
     uint256 public bar;             // Target Interest Rate [ray]
+    uint256 public pie;             // Maximum percentage this contract can hold of the pool's deposits
     uint256 public live = 1;
     uint256 public culled;
     uint256 public tic;             // Timestamp when the system is caged
@@ -205,6 +206,10 @@ contract DssDirectDepositAaveDai {
             require(data <= interestStrategy.getMaxVariableBorrowRate(), "DssDirectDepositAaveDai/above-max-interest");
 
             bar = data;
+        } else if (what == "pie") {
+            require(data <= RAY, "DssDirectDepositAaveDai/pie-greater-than-whole");
+
+            pie = data;
         } else revert("DssDirectDepositAaveDai/file-unrecognized-param");
 
         emit File(what, data);
@@ -237,7 +242,7 @@ contract DssDirectDepositAaveDai {
     }
 
     // --- Deposit controls ---
-    function _wind(uint256 amount) internal {
+    function _wind(uint256 amount, uint256 availableLiquidity) internal {
         // IMPORTANT: this function assumes Vat rate of this ilk will always be == 1 * RAY (no fees).
         // That's why this module converts normalized debt (art) to Vat DAI generated with a simple RAY multiplication or division
         // This module will have an unintended behaviour if rate is changed to some other value.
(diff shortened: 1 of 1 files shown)


src/DssDirectDepositAaveDai.sol
```diff
@@ -255,9 +260,14 @@ contract DssDirectDepositAaveDai {
...
         }
 
+
         require(int256(amount) >= 0, "DssDirectDepositAaveDai/overflow");
 
         uint256 scaledPrev = adai.scaledBalanceOf(address(this));
+        uint256 interestIndex = pool.getReserveNormalizedIncome(address(dai));
+        uint256 scaledNew = _add(scaledPrev, _rdiv(amount, interestIndex));
+
+        require(_rdiv(scaledNew, availableLiquidity) <= pie, "DssDirectDepositAaveDai/share-too-large");
```
hexonaut, line 270: This should probably compare the unscaled version to `availableLiquidity` as the scaled version is not the same as the amount of actual Dai available to the D3M. Also this needs to deal with the case where `availableLiquidity == 0` as this will be a division by zero which will fail, but maybe it's better to handle that explicitly.

iamchrissmith:
> So I think the code is mostly correct. There is just one fix I mentioned below.
> 
> That being said, I have two main problems with this approach:
> 
> 1. I don't think it's necessary. This bank run scenario is applicable to all users of the pool and so far I haven't seen this occur anywhere except in the case of the bZx hack from early 2020. I think there is maybe a case to be made for this in the event of a hack, but this leads me to point 2.
> 2. You can circumvent this limit with Flash Loans. Let's say we have a pool with 80M in outstanding loans and 20M in available liquidity. We set the pie to 50% so we can at most have D3M debt at the level of 10M with current conditions. Suddenly there is a hack and people scramble to withdraw all Dai in the pool. The pool is now 80M in outstanding loans and 0M in available liquidity. D3M will stop minting at this point with your new code addition, but I can circumvent that with the steps below:
> 
> Let's say I'm someone else with 10M in deposits in the pool, and I want to steal from the D3M. Steps are (in one tx):
> 
> 1. Flash loan 20M Dai.
> 2. Put 20M Dai into the aDai pool. `availableLiquidity` is now 20M with 80M OLs.
> 3. Withdraw your 10M deposit. `availableLiquidity` is now 10M with 80M OLs.
> 4. Call `d3m.exec()`. `availableLiquidity` is now 20M with 80M OLs.
> 5. Withdraw the original 20M flash loan deposit.
> 6. Repay flash loan.
> 
> With these steps I believe you can transfer all of the outstanding debt over to the D3M and circumvent the `availableLiquidity` limit.
> 
> In summary, I don't think this gives us any extra protection in the event of a bug, and I do not believe that a bank run scenario is likely to last outside of a hack due to the crazy high interest rates.

Thank you for looking at this @hexonaut . I missed the flashloan approach for getting around this and I agree, that completely circumvents this protection.  So it seems, at the end of the day, D3M will likely end up holding the aDai in the case of a bank run and governance needs to limit that risk through `line`.

PR #13, opened 2021-10-07
Allow temporarily turning off the module by setting the bar to zero

hexonaut:
Thanks @iamchrissmith for the suggestion. Seems like a handy feature to have.

Also, just adding in a minor fix for the `targetInterestRate` needing to be greater than the base borrow rate instead of 0.

### DIFF
--- a/src/DssDirectDepositAaveDai.sol
+++ b/src/DssDirectDepositAaveDai.sol
@@ -219,18 +219,19 @@ contract DssDirectDepositAaveDai {
 
     // --- Automated Rate targeting ---
     function calculateTargetSupply(uint256 targetInterestRate) public view returns (uint256) {
-        require(targetInterestRate > 0, "DssDirectDepositAaveDai/target-interest-zero");
+        uint256 base = interestStrategy.baseVariableBorrowRate();
+        require(targetInterestRate > base, "DssDirectDepositAaveDai/target-interest-zero");
         require(targetInterestRate <= interestStrategy.getMaxVariableBorrowRate(), "DssDirectDepositAaveDai/above-max-interest");
 
         // Do inverse calculation of interestStrategy
         uint256 targetUtil;
-        if (targetInterestRate > _add(interestStrategy.baseVariableBorrowRate(), interestStrategy.variableRateSlope1())) {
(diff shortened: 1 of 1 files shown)


src/DssDirectDepositAaveDai.t.sol
```diff
@@ -328,6 +328,24 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
     }
 
+    function test_bar_zero() public {
+        uint256 targetBorrowRate = _setRelBorrowTarget(7500);
+        deposit.reap();     // Clear out interest to get rid of rounding errors
+        assertEqInterest(getBorrowRate(), targetBorrowRate);
+
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(deposit));
+        assertGt(ink, 0);
+        assertGt(art, 0);
```
iamchrissmith, line 338: Why does the setup of this test result in `ink` and `art` = 0 when this test end up with some `ink` and `art`?
https://github.com/makerdao/dss-direct-deposit/blob/716821f80264158c193602ec70ea746a6ed4fd65/src/DssDirectDepositAaveDai.t.sol#L300-L308

hexonaut: It's a greater than. :) I miss those sometimes too.

hexonaut:
`targetInterestRate` needs to be greater than zero or you get a division by zero error on line 235 because `targetUtil` is zero.

Actually this value needs to be greater than `interestStrategy.baseVariableBorrowRate()`, so should probably fix that.

iamchrissmith:
> `targetInterestRate` needs to be greater than zero or you get a division by zero error on line 235 because `targetUtil` is zero.
> 
> Actually this value needs to be greater than `interestStrategy.baseVariableBorrowRate()`, so should probably fix that.

I think changing it to be `targetInterestRate >= interestStrategy.baseVariableBorrowRate()` makes sense. (Greater than or equal works, right?) 

I was thinking the line could be removed because of your change here `uint256 targetSupply = bar > 0 ? calculateTargetSupply(bar) : 0;` so it can never be called with `0`, but i remember now that this is `public` so we still need that protection.

iamchrissmith (approved):
lgtm

PR #14, opened 2021-10-07
Disable reap after cage to prioritize paying down debt vs fee collection

hexonaut:
Again thanks @iamchrissmith for this suggestion. We want to prioritize paying down debt instead of collecting fees during D3M cage.

### DIFF
--- a/src/DssDirectDepositAaveDai.t.sol
+++ b/src/DssDirectDepositAaveDai.t.sol
@@ -975,7 +975,7 @@ contract DssDirectDepositAaveDaiTest is DSTest {
         
         hevm.warp(block.timestamp + 1 days);    // Accrue some interest
 
-        // uncull should fail with error "DssDirectDepositAaveDai/no-reap-during-cage"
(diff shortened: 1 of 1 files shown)


src/DssDirectDepositAaveDai.t.sol
```diff
@@ -967,4 +967,15 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
     }
+    
+    function testFail_reap_caged() public {
+        _setRelBorrowTarget(7500);
+
+        deposit.cage();
+        
+        hevm.warp(block.timestamp + 1 days);    // Accrue some interest
+
+        // uncull should fail with error "DssDirectDepositAaveDai/no-reap-during-cage"
```
gbalabasquer, line 978:
```suggestion
        // reap should fail with error "DssDirectDepositAaveDai/no-reap-during-cage"
```

gbalabasquer (approved):
Minor error in a comment, but other than that LGTM

PR #15, opened 2021-10-08
Add a shouldExec helper function

hexonaut:
This is a helper function for the bot that will be calling `exec()` to have a repeatable call to determine whether the interest rate has deviated out of the acceptable tolerance.

For example calling `shouldExec(1 * RAY / 100)` will only return true if the Aave interest rate deviates 1% from the target `bar`.

### DIFF
--- a/src/DssDirectDepositAaveDai.sol
+++ b/src/DssDirectDepositAaveDai.sol
@@ -81,6 +81,18 @@ interface InterestRateStrategyLike {
     function variableRateSlope2() external view returns (uint256);
     function baseVariableBorrowRate() external view returns (uint256);
     function getMaxVariableBorrowRate() external view returns (uint256);
+    function calculateInterestRates(
+        address reserve,
+        uint256 availableLiquidity,
+        uint256 totalStableDebt,
+        uint256 totalVariableDebt,
+        uint256 averageStableBorrowRate,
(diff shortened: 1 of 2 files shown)


src/DssDirectDepositAaveDai.sol
```diff
@@ -236,6 +249,45 @@ contract DssDirectDepositAaveDai {
...
+        } else if (deviation > RAY) {
+            // Wind case
+            (uint256 Art,,, uint256 line,) = vat.ilks(ilk);
+            return _sub(line, Art*RAY) > RAD && (deviation - RAY) > interestRateTolerance;  // Allow for rounding errors, DC = full if 1 Dai away
+        } else {
+            // No change
+            return false;
+        }
+        
+        return deviation - RAY > interestRateTolerance;
```
iamchrissmith, line 288: how do we hit this case? I seem like the `if`s above will handle all of the branches?

gbalabasquer: Just realized you also mentioned it. Yeah this needs to be removed.

iamchrissmith (approved):
this looks good to me

hexonaut:
Going to hold off with this for now. It's probably better to leave this in a separate contract.

PR #16, opened 2021-10-12
Add functions to support the gem join standard interface

hexonaut:
This will enable usage of the `DssExecLib.addNewCollateral();` as well as make it friendly with the ilk registry and such.


talbaneth (approved):
LGTM

PR #21, opened 2021-11-04
ChainSecurity Audit Fixes

hexonaut:
* Cache `variableRateSlope1` to save gas


iamchrissmith (approved):
lgtm

gbalabasquer:
As this is coming as a change after deployment I'd suggest we tag and release the version we have in production right now.

PR #22, opened 2021-11-12
Add Helper Contract for Keeper Bots

hexonaut:
Instead of just submitting a tx every X duration this will allow keeper bots to check if a threshold has been passed which will make things much more efficient.

### DIFF
--- a/src/helper/DirectHelper.sol
+++ b/src/helper/DirectHelper.sol
@@ -88,7 +88,7 @@ contract DirectHelper {
         bytes32 ilk = direct.ilk();
         InterestRateStrategyLike interestStrategy = InterestRateStrategyLike(direct.interestStrategy());
 
-        (uint256 daiDebt,) = vat.urns(ilk, address(this));
+        (uint256 daiDebt,) = vat.urns(ilk, address(direct));
         uint256 _bar = direct.bar();
         if (_bar == 0) {
             return daiDebt > 0;     // Always attempt to close out if we have debt remaining
--- a/src/helper/DirectHelper.sol
+++ b/src/helper/DirectHelper.sol
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2021 Dai Foundation
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity 0.6.12;
+
+interface VatLike {
+    function ilks(bytes32) external view returns (uint256, uint256, uint256, uint256, uint256);
+    function urns(bytes32, address) external view returns (uint256, uint256);
+}
+
+interface DirectLike {
+    function vat() external view returns (address);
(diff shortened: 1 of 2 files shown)


src/helper/DirectHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        address _direct,
+        uint256 interestRateTolerance
+    ) external view returns (bool) {
+        // IMPORTANT: this function assumes Vat rate of this ilk will always be == 1 * RAY (no fees).
+        // That's why this module converts normalized debt (art) to Vat DAI generated with a simple RAY multiplication or division
+        // This module will have an unintended behaviour if rate is changed to some other value.
+
+        DirectLike direct = DirectLike(_direct);
+        VatLike vat = VatLike(direct.vat());
+        TokenLike dai = TokenLike(direct.dai());
```
iamchrissmith, line 84: I think this can be an `address`
```suggestion
        address dai = direct.dai();
```

src/helper/DirectHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        (,, uint256 currVarBorrow) = interestStrategy.calculateInterestRates(
+            adai,
+            dai.balanceOf(adai),
+            stableDebt.totalSupply(),
+            variableDebt.totalSupply(),
+            0,
+            0
+        );
```
iamchrissmith, line 104: in D3M for `LendingPoolLike` the comments indicate that we could get the current variable rate using `direct.pool.getReserveData(address(dai))`

I think this makes things similar and seems to work: https://github.com/makerdao/dss-direct-deposit/pull/23

Let me know if there is some Aave implementation detail I'm missing (I didn't really find this section in their docs, so kind of piecing things together).

Gas diff: 1472009 -> 1274197

src/helper/DirectHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+
+        uint256 deviation = _rdiv(currVarBorrow, _bar);
+        if (deviation < RAY) {
+            // Unwind case
+            return daiDebt > 0 && (RAY - deviation) > interestRateTolerance;
```
iamchrissmith, line 109: should this be `>= interestRateTolerance`?

src/helper/DirectHelper.sol
```diff
@@ -0,0 +1,120 @@
...
+        } else if (deviation > RAY) {
+            // Wind case
+            (uint256 Art,,, uint256 line,) = vat.ilks(ilk);
+            return Art*RAY < line && (deviation - RAY) > interestRateTolerance;
```
iamchrissmith, line 113: should this be `>= interestRateTolerance`?

iamchrissmith: This does not calculate whether or not there is room in the debt ceiling to succeed in the `wind` so if we are 1 wei below `line` and outside our tolerance the bot would call `exec` but would revert, right? Do you think it is worth calling `calculateTargetSupply` to see if there is room in the `line` to wind?

iamchrissmith (changes requested):
Couple of comments and a suggestion on the current rate fetch. let me know what you think

src/helper/DirectHelper.sol
hexonaut: It doesn't really matter tbh. This interest tolerance is just an arbitrary number picked by the bot operator.

src/helper/DirectHelper.sol
hexonaut: Yeah I thought about this case, and I think it will exactly max out whenever it can, so it's really unlikely to be within some small delta of the debt ceiling.

Anyways, this is not a high stakes contract to get perfect. Worst case we burn a bit of gas and redeploy.

iamchrissmith: oh you're right it takes either the amount to add to change the rate or the delta to fill up the `line` 👍🏻

iamchrissmith (dismissed):
lgtm (after tests are fixed)

hexonaut:
The failing test is unrelated to this PR. This test need to be fixed, but will do it in another PR.

PR #23, opened 2021-11-13
use the pools rate instead of calculating it from the strategy

iamchrissmith:
Change to use the pool instead of the strategy to get the current `currVarBorrow` rate. Seems to save some gas


hexonaut (approved):
Ah yes much better.

PR #33, opened 2022-03-22
Direct helper no longer used

hexonaut:
The helper contract is no longer being used, and has been replaced by [the dss-cron contract](https://github.com/makerdao/dss-cron/blob/master/src/AaveDirectJob.sol).


iamchrissmith (approved):

naszam:
I think direct helper coverage in tests (`test_shouldExec`) could be removed as well.

hexonaut:
Good catch. Totally forgot to run the tests.

hexonaut:
Should be good now.

naszam (approved):

PR #34, opened 2022-03-22
Update to Eip 4626

iamchrissmith:
updating the language to better reflect EIP-4626


src/DssDirectDepositHub.sol
```diff
@@ -215,7 +214,7 @@ contract DssDirectDepositHub {
         // This module will have an unintended behaviour if rate is changed to some other value.
 
         address end;
-        uint256 gemBalance = pool.gemBalanceOf();
+        uint256 gemBalance = pool.maxRedeem();
```
hexonaut, line 217: This needs to be reworked to use `convertToAssets(balanceOf())` to get the total position size in DAI, then the amount to withdraw is constrained by `maxWithdraw()` to get `availableLiquidity`.

Would also stick with language `asset` for the language instead of `gem`. IE `gemBalance` -> `assetBalance`.

src/DssDirectDepositHub.sol
```diff
@@ -295,7 +294,7 @@ contract DssDirectDepositHub {
     function exec(bytes32 ilk_) external {
         Ilk memory ilk = ilks[ilk_];
 
-        uint256 availableLiquidity = dai.balanceOf(address(ilk.gem));
+        uint256 availableAssets = dai.balanceOf(address(ilk.gem));
```
hexonaut, line 297: This should be changed to `maxWithdraw()` and merged into `_unwind()`.

hexonaut (commented):
Heading along the right track. I would remove all references to `gem` and abstract those into the pool. IE `gem.transfer(who, ???)` -> `pool.transfer(who, sharesAmount)`, etc

Will continue review after comments are addressed as I think they will require a few somewhat significant changes.

src/DssDirectDepositHub.sol
iamchrissmith: I think this should be shares. Gem at this point was the adai balance.  Let me know if that makes sense to you

iamchrissmith:
@hexonaut I made some of the updates you suggested after rebasing to get the helper changes out.
 
Main differences between your suggestions and the changes:
- instead of `convertToAssets(balanceOf())` i used `maxWithdraw`. It seems like the EIP says that should return the asset/dai balance so should avoid needing to convert the share balance to assets.
- I did not merge `uint256 availableAssets = maxWithdraw` from `exec` to `unwind` because we need it for the calc supplies and `wind` call too (i think)

let me know if that makes sense and if i messed any of the other conversions up.

hexonaut:
Will do another review in a bit. Getting on a plane. Answering some questions first:

> instead of `convertToAssets(balanceOf())` i used `maxWithdraw`. It seems like the EIP says that should return the asset/dai balance so should avoid needing to convert the share balance to assets.

This was originally my thought as well, but `maxWithdraw` is not necessarily the balance of the user. It is the most they can withdraw which may or may not be the balance. For example, in the Aave this could be constrained if there is not enough DAI liquidity available in the pool to withdraw your total balance. The definition says the highest amount you can withdraw without causing a revert.

> I did not merge uint256 availableAssets = maxWithdraw from exec to unwind because we need it for the calc supplies and wind call too (i think)

Ok, will check this out in detail when I get a chance.

iamchrissmith:
@hexonaut for:

> maxWithdraw is not necessarily the balance of the user.

That is correct, but for the exec determination don't we want all available Dai to be withdrawn, not just the balance held by the D3M pool contract?  For the `quit` function we want the contract's shares/balance so i used `shareBalance` there. is that make sense you think?

hexonaut:
> That is correct, but for the exec determination don't we want all available Dai to be withdrawn, not just the balance held by the D3M pool contract? For the `quit` function we want the contract's shares/balance so i used `shareBalance` there. is that make sense you think?

For `quit` I think it's fine, but in the case of the regular `_unwind()` we need to determine how much of a position we have to ensure it is always over-collateralized when extracting out interest. This is why we need to get the total value of the position and compare it to the outstanding debt in DAI. Anything above this in the position's total value can be safely remove and pocketed as interest earned.

iamchrissmith:
> > That is correct, but for the exec determination don't we want all available Dai to be withdrawn, not just the balance held by the D3M pool contract? For the `quit` function we want the contract's shares/balance so i used `shareBalance` there. is that make sense you think?
> 
> For `quit` I think it's fine, but in the case of the regular `_unwind()` we need to determine how much of a position we have to ensure it is always over-collateralized when extracting out interest. This is why we need to get the total value of the position and compare it to the outstanding debt in DAI. Anything above this in the position's total value can be safely remove and pocketed as interest earned.
> 
> Ex) Let's say the total position + interest is 101m and we have 100m in DAI debt. We can safely extract 1m of DAI as interest and put it in the surplus buffer. Now let's say the pool is utilized such that `maxWithdraw()` is limited to 50m. We can still safely extract 1m of that as interest, but how do we determine this? 50m < 100m.

ok, yes i see what you are saying. I was looking at the `availableAssets` in exec which should be `maxWithdraw` but i see that you are talking about `assetBalance` in `_unwind` which I agree, I have wrong so will update. thanks!

fixed: https://github.com/makerdao/dss-direct-deposit/pull/34/commits/8b76a0819c7db70ec1032902e6dd839cb502d909

src/DssDirectDepositHub.sol
```diff
@@ -344,13 +343,13 @@ contract DssDirectDepositHub {
...
+        uint256 assetBalance = ilk.pool.convertToAssets(ilk.pool.shareBalance());
         (, uint256 daiDebt) = vat.urns(ilk_, address(ilk.pool));
-        if (gemBalance > daiDebt) {
-            uint256 fees = gemBalance - daiDebt;
-            uint256 availableLiquidity = dai.balanceOf(address(ilk.gem));
-            if (fees > availableLiquidity) {
-                fees = availableLiquidity;
+        if (assetBalance > daiDebt) {
+            uint256 fees = assetBalance - daiDebt;
+            uint256 availableAssets = ilk.pool.assetBalance();
```
hexonaut, line 350: This should be `maxWithdraw()` instead of `assetBalance()`.

hexonaut (commented):
Ok looking good now. Can this current interface be added as a separate interface file for us to implement? After that I think this can be merged and work start on pool-specific connectors.

PR #35, opened 2022-03-24
Remove `dss-interfaces` in tests

naszam:
- [x] Remove `dss-interfaces` in tests to reduce dependency maintenance
- [x] Use Static Interfaces in tests, inheriting contract interfaces to be expanded where needed.

### DIFF
--- a/src/DssDirectDepositAaveDai.t.sol
+++ b/src/DssDirectDepositAaveDai.t.sol
@@ -17,10 +17,9 @@
 pragma solidity 0.6.12;
 
 import "ds-test/test.sol";
-import "dss-interfaces/Interfaces.sol";
 import "ds-value/value.sol";
 
-import {DssDirectDepositAaveDai} from "./DssDirectDepositAaveDai.sol";
+import "./DssDirectDepositAaveDai.sol";
 import {DirectDepositMom} from "./DirectDepositMom.sol";
 
 interface Hevm {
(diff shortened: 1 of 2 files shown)


src/DssDirectDepositAaveDai.t.sol
```diff
@@ -109,18 +122,18 @@ contract DssDirectDepositAaveDaiTest is DSTest {
...
-        adai = DSTokenAbstract(0x028171bCA77440897B824Ca71D1c56caC55b68A3);
-        stkAave = DSTokenAbstract(0x4da27a545c0c5B758a6BA100e3a049001de870f5);
-        dai = DaiAbstract(0x6B175474E89094C44Da98b954EedeAC495271d0F);
-        daiJoin = DaiJoinAbstract(0x9759A6Ac90977b93B58547b4A71c78317f391A28);
-        interestStrategy = InterestRateStrategyLike(0xfffE32106A68aA3eD39CcCE673B646423EEaB62a);
-        rewardsClaimer = RewardsClaimerLike(0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5);
+        pool = LendingPoolAbstract(0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9);
+        adai = TokenLike(0x028171bCA77440897B824Ca71D1c56caC55b68A3);
+        stkAave = TokenLike(0x4da27a545c0c5B758a6BA100e3a049001de870f5);
+        dai = TokenLike(0x6B175474E89094C44Da98b954EedeAC495271d0F);
```
brianmcmichael, line 131: Head's up `DaiAbstract` and `DSTokenAbstract` have different interfaces. The cast will probably work here, but we might want to be explicit here in case tests get expanded on.

naszam: yeah, was wondering the same, just followed the code pattern for now, as it uses `TokenLike` also for Dai.
I agree on splitting those interfaces for compatibility in case of future test expansions.

naszam: done in bd15e60

src/DssDirectDepositAaveDai.t.sol
```diff
@@ -29,33 +28,55 @@ interface Hevm {
     function load(address,bytes32) external view returns (bytes32);
 }
 
-interface AuthLike {
-    function wards(address) external returns (uint256);
+interface GemAbstract {
```
gbalabasquer, line 31: Why not to call them `Like` everywhere? and maybe avoid the inheritance just duplicate the functions whenever is necessary.
using: `import({...} from '');` to import just the contract to test will avoid the issue of using the same naming for the interfaces.

naszam: yeah, this approach was used in `vote-delegate`.
The current one make use of existing code interfaces expanding them were needed, to reduce code size, favouring reusability.
I see your point though, and I think it's more elegant overall.

naszam: done in 06bd9e9

brianmcmichael:
This PR may be premature due to the work going on in the `modularize` branch.

It may be better to just scrap this and do something similar that we can merge into that stream.

naszam:
> This PR may be premature due to the work going on in the `modularize` branch.
> 
> It may be better to just scrap this and do something similar that we can merge into that stream.

checked with @iamchrissmith, following the discussion in chat, the reason to branch off master was to remove `dss-interfaces` as dependency and reflect those changes on `dss-exec-lib` (in particular in `rwa-dc` pr) to reduce `dss-interfaces` dependency maintenance, as every new entry in `dss-interfaces` would need to go into `dss-direct-deposit` before being upgraded in `dss-exec-lib` as well.

iamchrissmith:
> > This PR may be premature due to the work going on in the `modularize` branch.
> > It may be better to just scrap this and do something similar that we can merge into that stream.
> 
> checked with @iamchrissmith, following the discussion in chat, the reason to branch off master was to remove `dss-interfaces` as dependency and reflect those changes on `dss-exec-lib` (in particular in `rwa-dc` pr) to reduce `dss-interfaces` dependency maintenance, as every new entry in `dss-interfaces` would need to go into `dss-direct-deposit` before being upgraded in `dss-exec-lib` as well.

I didn't want to block @naszam on this work since modularize will require discussions. I'll rebase once these changes are in and update what i need to in my branch to clean up the conflicts.

iamchrissmith (approved):
lgtm

gbalabasquer (approved):
LGTM

PR #37, opened 2022-03-25
Modularize upd 325

brianmcmichael:
Modifications I was able to make from a review pass. Will bring up some more structural stuff elsewhere.


iamchrissmith (approved):
thank you!

PR #38, opened 2022-04-07
feat: Add missing functionality to `D3MMapleV1DaiPool`, contracts compiling

lucas-manuel:
Address some TODOs to get initial structure in place to begin testing.

### DIFF
--- a/src/pools/D3MMapleV1DaiPool.sol
+++ b/src/pools/D3MMapleV1DaiPool.sol
@@ -20,10 +20,13 @@ import "../bases/D3MPoolBase.sol";
 
 interface PoolLike is TokenLike {
     function deposit(uint256 amount) external;
+    function depositDate(address) external view returns (uint256);
     function intendToWithdraw() external;
     function liquidityCap() external view returns (uint256);
     function liquidityLocker() external view returns (address);
+    function lockupPeriod() external view returns (uint256);
     function principalOut() external view returns (uint256);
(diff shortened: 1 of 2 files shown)


src/pools/D3MMapleV1DaiPool.sol
```diff
@@ -59,16 +70,24 @@ contract D3MMapleV1DaiPool is D3MPoolBase {
...
     function deposit(uint256 amt) external override auth {
-        // TODO confirm deposit is in units of DAI
-        pool.deposit(amt);
-        // TODO: emit deposit event if we decide to leave it in base
+        pool.deposit(amt);  // Deposit DAI, recieve LP tokens 1-to-1
+
+        // TODO: Emit deposit event if we decide to leave it in base
+    }
+
+    function intendToWithdraw() external auth {
```
hexonaut, line 78: I added the `intendToWithdraw` logic into the Plan. The idea is that governance will set the loan amount to 200m target, then the Hub logic winds up the position as room becomes available. When we want to close it out we tell the Plan to lower to 100m or something which will trigger the `intendToWithdraw` at that point if the current loan is larger than 100m. Depending on where we are in the cooldown periods this may or may not require a keeper.

You can see the logic I outlined in the Plan. Would prefer to keep it in there if possible.

hexonaut:
Going to merge and take over.

PR #41, opened 2022-04-26
Add valid target check to exec() normal path

talbaneth:


hexonaut (approved):

gbalabasquer (approved):
LGTM

iamchrissmith (approved):
This looks good to me. I am fine with this approach but was wondering whether we should unwind everything if the plan we used is no longer valid.  We could always rewind it once it is fixed.  I'm not sure what the risk is of having our position sit there, but seems it could be safer to unwind rather than just reverting here.  We can get there other ways, so like I said, this lgtm to too

iamchrissmith:
> > This looks good to me. I am fine with this approach but was wondering whether we should unwind everything if the plan we used is no longer valid. We could always rewind it once it is fixed. I'm not sure what the risk is of having our position sit there, but seems it could be safer to unwind rather than just reverting here. We can get there other ways, so like I said, this lgtm to too
> 
> That's actually possible calling first `cage` and then `exec`. It's true we could replace the require condition and put it in the if conditional for unwind. I think I'm fine both ways.

100%, but that was the case before this change too.  Good either way

talbaneth:
> > > This looks good to me. I am fine with this approach but was wondering whether we should unwind everything if the plan we used is no longer valid. We could always rewind it once it is fixed. I'm not sure what the risk is of having our position sit there, but seems it could be safer to unwind rather than just reverting here. We can get there other ways, so like I said, this lgtm to too
> > 
> > 
> > That's actually possible calling first `cage` and then `exec`. It's true we could replace the require condition and put it in the if conditional for unwind. I think I'm fine both ways.
> 
> 100%, but that was the case before this change too. Good either way

Well before this change you could perform the exec if you front run the `cage` operation.
I opted for revert mostly because I didn't want to bypass the existing caging mechanism that calls `pool.cage()` and sets off the `tic` counter.
Also this way the change was simpler.

PR #43, opened 2022-05-02
add Natspec comments to base Pool, Plan and Hub contracts

iamchrissmith:
Adding Natspec comments to contracts

### DIFF
--- a/src/DssDirectDepositHub.sol
+++ b/src/DssDirectDepositHub.sol
@@ -75,7 +75,7 @@ contract DssDirectDepositHub {
     /**
         @notice Maps address that have permission in the Pool
         @dev 1 = allowed, 0 = no permission
-        @return authorization, 1 or 0
+        @return authorization 1 or 0
     */
     mapping (address => uint256) public wards;
 
@@ -150,8 +150,6 @@ contract DssDirectDepositHub {
 
     /**
         @dev sets msg.sender as authed.
-        Gives infinite approval for Dai on the DaiJoin contract.
-        Hopes the DaiJoin in the Vat.
         @param vat_     address of the DSS vat contract
         @param daiJoin_ address of the DSS Dai Join contract
     */
@@ -338,11 +336,11 @@ contract DssDirectDepositHub {
 
     /**
         @notice Main function for updating a D3M position.
-        Determines the current state and either winds or unwinds as necessary
+        Determines the current state and either winds or unwinds as necessary.
         @dev Winding the target position will be constrained by the Ilk debt ceiling,
         the overall DSS debt ceiling and the maximum deposit by the pool.
         Unwinding the target position will be constrained by the number of assets available
-        to be withdrawn from the pool
+        to be withdrawn from the pool.
         @param ilk bytes32 of the D3M ilk name
     */
     function exec(bytes32 ilk) external {
@@ -442,10 +440,10 @@ contract DssDirectDepositHub {
 
     /**
         @notice Allow Users to return vat gem for Pool Shares.
-        This will only occur during Global Settlement when users receive collateral for their Dai
+        This will only occur during Global Settlement when users receive collateral for their Dai.
         @param ilk bytes32 of the D3M ilk name
-        @param usr  address that should receive the shares from the pool
-        @param wad  amount of gems that the msg.sender is returning. Note: this could be different than the number of shares usr will receive
+        @param usr address that should receive the shares from the pool
+        @param wad amount of gems that the msg.sender is returning. Note: this could be different than the number of shares usr will receive
     */
     function exit(bytes32 ilk, address usr, uint256 wad) external {
         require(wad <= 2 ** 255, "DssDirectDepositHub/overflow");
@@ -511,8 +509,8 @@ contract DssDirectDepositHub {
     /**
         @notice Rollback Write-off (cull) if General Shutdown happened.
         This function is required to have the collateral back in the vault so it can be taken by End module
-        and eventually be shared to DAI holders (as any other collateral) or maybe even unwinded
-        @dev This pulls gems from the pool and reopens the urn with the gem amount of ink/art
+        and eventually be shared to DAI holders (as any other collateral) or maybe even unwinded.
+        @dev This pulls gems from the pool and reopens the urn with the gem amount of ink/art.
         @param ilk bytes32 of the D3M ilk name
     */
     function uncull(bytes32 ilk) external {
@@ -537,7 +535,7 @@ contract DssDirectDepositHub {
         the urn to the recipient.
         @dev If called while not culled, it will require who to hope on the Hub contract in the Vat
         @param ilk bytes32 of the D3M ilk name
-        @param who  address of who will receive the shares and possibly the urn
+        @param who address of who will receive the shares and possibly the urn
     */
     function quit(bytes32 ilk, address who) external auth {
         require(vat.live() == 1, "DssDirectDepositHub/no-quit-during-shutdown");
--- a/src/DssDirectDepositHub.sol
+++ b/src/DssDirectDepositHub.sol
@@ -459,7 +459,7 @@ contract DssDirectDepositHub {
         @notice Shutdown a pool.
         This starts the countdown to when the debt can be written off (cull).
         Once called, subsequent calls to `exec` will unwind as much of the position as possible.
-        @dev As long as the pool returns true for `validTarget()` this requires msg.sender be auth'ed
+        @dev msg.sender must be authorized
         @param ilk bytes32 of the D3M ilk name
     */
     function cage(bytes32 ilk) external auth {
--- a/src/DssDirectDepositHub.sol
+++ b/src/DssDirectDepositHub.sol
@@ -64,18 +64,42 @@ interface TokenLike {
     function approve(address, uint256) external returns (bool);
 }
 
+/**
+    @title D3M Hub
+    @notice This is the main D3M contract and is responsible for winding/unwinding pools,
+    interacting with DSS and tracking the plans and pools and their states.
+*/
 contract DssDirectDepositHub {
 
     // --- Auth ---
+    /**
+        @notice Maps address that have permission in the Pool
+        @dev 1 = allowed, 0 = no permission
+        @return authorization, 1 or 0
+    */
     mapping (address => uint256) public wards;
+
+    /**
+        @notice Makes an address authorized to perform auth'ed functions
+        @dev msg.sender must be authorized
+        @param usr address to be authorized
+    */
     function rely(address usr) external auth {
         wards[usr] = 1;
         emit Rely(usr);
     }
+
+    /**
+        @notice De-authorizes an address from performing auth'ed functions
+        @dev msg.sender must be authorized
+        @param usr address to be de-authorized
+    */
     function deny(address usr) external auth {
         wards[usr] = 0;
         emit Deny(usr);
     }
+
+    /// @notice Modifier will revoke if msg.sender is not authorized
     modifier auth {
         require(wards[msg.sender] == 1, "DssDirectDepositHub/not-authorized");
         _;
@@ -89,6 +113,14 @@ contract DssDirectDepositHub {
     address      public           vow;
     EndLike      public           end;
 
+    /**
+        @notice Tracking struct for each of the D3M ilks
+        @param pool   Contract to access external pool and hold balances
+        @param plan   Contract used to calculate target debt
+        @param tau    Time until you can write off the debt [sec]
+        @param culled Debt write off triggered (1 or 0)
+        @param tic    Timestamp when the pool is caged
+    */
     struct Ilk {
         D3MPoolLike pool;   // Access external pool and holds balances
         D3MPlanLike plan;   // How we calculate target debt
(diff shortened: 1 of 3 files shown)


src/pools/D3MPoolBase.sol
```diff
@@ -70,26 +114,90 @@ abstract contract D3MPoolBase {
...
 
+    /**
+        @notice Some external pools require manually accruing fees/interest
+        @dev Used in Hub.exec() and Hub.reap(), can be empty if external pool does not require
+    */
     function accrueIfNeeded() external virtual;
 
+    /**
+        @notice Balance of Assets this pool "owns".
+        Should be Deposit + Interest/Fees earned
```
hexonaut, line 165: In general it will not be specifically just deposit + interest. It could subtract out losses as well. I would just leave it open instead of defining what this should equal.

src/pools/D3MPoolBase.sol
```diff
@@ -70,26 +114,90 @@ abstract contract D3MPoolBase {
...
+        @dev This could be greater than the amount the pool can withdraw due to lack of liquidity
+        @return uint256 Number of Assets in Dai [wad]
+    */
     function assetBalance() external view virtual returns (uint256);
 
+    /**
+        @notice Maximum number of Assets the pool could withdraw at present.
+        Should be the minimum between:
+            - Deposit + Interest/Fees earned
+            - the available liquidity in the pool
```
hexonaut, line 175: Same here. This is a bit too prescriptive for the base contract imo. There could be any number of constraints limiting this depending on what the target protocol is.

hexonaut (changes requested):

src/DssDirectDepositHub.sol
```diff
@@ -376,15 +459,18 @@ contract DssDirectDepositHub {
         emit Exit(ilk_, usr, wad);
     }
 
-    // --- Shutdown ---
+    /**
+        @notice Shutdown a pool.
+        This start the countdown to when the debt can be written off (cull).
```
iamchrissmith, line 464:
```suggestion
        This starts the countdown to when the debt can be written off (cull).
```

src/pools/D3MPoolBase.sol
iamchrissmith, line 165:
```suggestion
```

src/DssDirectDepositHub.sol
```diff
@@ -113,6 +144,14 @@ contract DssDirectDepositHub {
     event Quit(bytes32 indexed ilk, address indexed usr);
     event Exit(bytes32 indexed ilk, address indexed usr, uint256 amt);
 
+    /**
+        @dev sets msg.sender as authed
+        Allows joining/exiting Dai by:
+           - giving infinite approval for Dai on the DaiJoin contract
+           - hoping the DaiJoin in the Vat
```
talbaneth, line 151: These seem a bit too detailed to me.

src/DssDirectDepositHub.sol
```diff
@@ -113,6 +144,14 @@ contract DssDirectDepositHub {
...
+        @param vat_     address of the DSS vat contract
```
talbaneth, line 152: Regardind `address` or `Address` (whether to capitalize the first letter of the description) - this is uninified throughut the documentation, suggest to align everywhere.

src/DssDirectDepositHub.sol
```diff
@@ -163,6 +216,16 @@ contract DssDirectDepositHub {
         emit File(ilk, what, data);
     }
 
+    /**
+        @notice update plan or pool addresses for D3M ilk
+        @dev Requirements:
+        - msg.sender must be authorized
+        - vat is live
+        - the pool is live
```
talbaneth, line 224: Also here, seems a bit too detailed to me, we don't need to describe all the validations.

src/DssDirectDepositHub.sol
```diff
@@ -274,6 +337,17 @@ contract DssDirectDepositHub {
...
     }
 
+    /**
+        @notice Main function for updating a D3M position.
+        Determines the current state and either winds or unwinds as necessary:
+        - MCD Caged: will unwind as much of the D3M position as possible
+        - Pool Caged/Not Culled: will unwind as much of the D3M position as possible based on the urn debt
+        - Pool Caged/Culled: will unwind as much of the D3M position as possible based on the Pool gem balance
+        - Normal Operation: Will use the D3M Plan to determine how big the desired position is and then either wind or unwind to that amount of Dai debt.
+        @dev Winding the target position will be constrained by the Ilk debt ceiling and the overall DSS debt ceiling.
```
talbaneth, line 347: Also here, this seems too much info (and will be hard to maintain when/if the logic changes).

talbaneth (commented):
Nice work for explaining some of this complex stuff.
See comments.

src/DssDirectDepositHub.sol
```diff
@@ -366,8 +444,13 @@ contract DssDirectDepositHub {
         }
     }
 
-    // --- Allow DAI holders to exit during global settlement ---
-    // wad: should be amount of gems, this could be different than the share tokens
+    /**
+        @notice Allow Users to return vat gem for Pool Shares.
```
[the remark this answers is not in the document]
iamchrissmith: do you think I should leave it as `return vat gems`? I guess since we removed shares because some pools might not have them, this could just be burning vat.gems for nothing or something more exotic than an ERC-20 share.

src/pools/D3MPoolBase.sol
```diff
@@ -70,26 +114,87 @@ abstract contract D3MPoolBase {
...
+        @notice Function that will withdraw assets (Dai) from the external pool.
+        @dev If the external pool requires a different amount be passed in the
+        conversion should occur here as the Hub passes Dai [wad] amounts.
+        @param amt Amount in Dai/asset terms that we want to withdraw
+    */
     function withdraw(uint256 amt) external virtual;
 
+    /**
+        @notice Function that will transfer shares.
+        @dev If the external pool/shares contract requires a different amount be passed in the
```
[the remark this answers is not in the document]
iamchrissmith: Yeah I'm still unclear how pools without shares are going to work... In this case, what are we transferring during `quit` and `exit`? I feel like the concept of a share (i.e. representation that you have deposited something in the pool) will necessarily exist even if its not an ERC-20 contract.  Either way, do you have a suggestion for how I remove "share" from the transfer functions? Feels weird to say: "will transfer something if the pool implements an underlying transferable concept"

src/plans/D3MPlanBase.sol
```diff
@@ -16,38 +16,79 @@
...
 
     // --- Events ---
     event Rely(address indexed usr);
     event Deny(address indexed usr);
     event Disable();
 
+    /**
+        @notice Sets immutable values, handles auth
+        @dev sets msg.sender as authed
+        @param dai_ address of the Dai contract. Assigned to dai
```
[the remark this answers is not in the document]
talbaneth: I actually mean about the `Assigned to dai` part (I think the removal you did is also good).

src/DssDirectDepositHub.sol
talbaneth: Not critical but I see the detailed explanation is still there (just not in bullets).
Let me know if you prefer to leave as is.

src/DssDirectDepositHub.sol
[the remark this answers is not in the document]
talbaneth: Actually I don't see an immediate better replacement for the explanations that use `shares` so I guess let's leave as is.

src/pools/D3MPoolBase.sol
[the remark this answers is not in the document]
talbaneth: As mentioned I don't see an immediate better replacement for the explanations that use shares so I guess let's leave as is.

talbaneth:
Almost all my comments handled, see small comments.
If possible please do another sweep about the capitalization of the first letter of explanations. It is still not unified.
For example:
```
        @param dst **address** that should receive the shares
        @param amt **Amount** in Gem terms that we want to withdraw
```

iamchrissmith:
@talbaneth I had to force push because of squash merging modularize, but I think I caught all your comments.

src/plans/ID3MPlan.sol
```diff
@@ -16,11 +16,28 @@
 
 pragma solidity 0.6.12;
 
+/**
+    @title D3M Plan Interface
+    @notice Plan contracts are the contract that the Hub uses to determine how much to change its position
```
talbaneth, line 21:
```suggestion
    @notice Plan contracts are contracts that the Hub uses to determine how much to change its position
```

src/plans/ID3MPlan.sol
```diff
@@ -16,11 +16,28 @@
...
+*/
 interface ID3MPlan {
     event Disable();
 
-    // --- Plan Functions ---
+    /**
+        @notice Determines what the position should be based on current assets and the
```
talbaneth, line 27: Why is there a line break in the middle of this comment but in other (sometimes) longer comments not?

src/plans/ID3MPlan.sol
```diff
@@ -16,11 +16,28 @@
...
+        @return uint256 target assets the Hub should wind or unwind to. Dai [wad] denomination
+    */
     function getTargetAssets(uint256 currentAssets) external view returns (uint256);
+
+    /// @notice Reports whether the plan is active
     function active() external view returns (bool);
+
+    /**
+        @notice Disables the plan so that it would instruct the Hub to unwind its entire position.
+        @dev It is up to the specific implentation exactly what this means and if it is reversible
```
talbaneth, line 39:
```suggestion
        @dev It is up to the specific implementation exactly what this means and if it is reversible
```

src/plans/ID3MPlan.sol
```diff
@@ -16,11 +16,28 @@
...
+        base implementation is permissionless.
```
talbaneth, line 40: There is no base implementation now, better say something like "Implementation should be permissioned".

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
 
 pragma solidity 0.6.12;
 
+/**
+    @title D3M Pool Interface
+    @notice Pool contracts are the contract that the Hub uses to standardize interactions with external Pools
```
talbaneth, line 21:
```suggestion
    @notice Pool contracts are contracts that the Hub uses to standardize interactions with external Pools
```

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+    @dev This contract will hold any balance provided by the external pool as well as the balance in the Vat.
```
talbaneth, line 22: "this contract" => "implementing contracts"?

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+    This contract aims to use EIP-4626 guidelines for assets/shares/maxWithdraw etc.
```
talbaneth, line 23:
```suggestion
    This interface aims to use EIP-4626 guidelines for assets/shares/maxWithdraw etc.
```

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+*/
 interface ID3MPool {
+    /**
+        @notice Function that will deposit assets (Dai) in the external pool.
```
talbaneth, line 27: Subjective but I would drop the "Function that will.." prefix everywhere.

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+        @dev If the external pool requires a different amount be passed in the
```
talbaneth, line 28:
```suggestion
        @dev If the external pool requires a different amount to be passed in, the
```

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+        conversion should occur here as the Hub passes Dai [wad] amounts.
+        @param amt amount in asset (Dai) terms that we want to deposit
+    */
     function deposit(uint256 amt) external;
```
talbaneth, line 32: In other places you write msg.sender must be authorized, but not here.
Relevant also for `withdraw()`, `transfer()`, `transferAll()`.

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+
+    /**
+        @notice Balance of assets this pool "owns".
+        @dev This could be greater than the amount the pool can withdraw due to lack of liquidity
+        @return uint256 number of assets in Dai [wad]
+    */
     function assetBalance() external view returns (uint256);
+
+    /**
+        @notice Maximum number of assets the pool could withdraw at present.
```
talbaneth, line 70:
```suggestion
        @notice Maximum number of assets the pool could deposit at present.
```

src/pools/ID3MPool.sol
```diff
@@ -16,15 +16,82 @@
...
+        @return uint256 number of assets in Dai [wad]
+    */
     function maxDeposit() external view returns (uint256);
+
+    /**
+        @notice Maximum number of assets the pool could deposit at present.
```
talbaneth, line 76:
```suggestion
        @notice Maximum number of assets the pool could withdraw at present.
```

talbaneth (changes requested):
I think this can be the last review from my side, see comments.

General stuff:
* Some comments end with a period and some don't, suggest to do a sweep and unify behaviour to whatever we like.
* Comments length before line breaks also seem to vary. Suggest to look at that as well.

talbaneth (approved):
LGTM

hexonaut (approved):

PR #47, opened 2022-05-10
Solc upgrade

iamchrissmith:


src/plans/D3MAaveDaiPlan.sol
```diff
@@ -0,0 +1,198 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.8.13;
```
naszam, line 17: wondering if this should be `>=0.6.12` and deployed with 0.8.13 instead, as I think it could break compilation as dependency of `dss-exec-lib` in tests (`DssAction.t.sol`) for the `bar` action.
cc: @brianmcmichael

Makefile
```diff
@@ -2,5 +2,5 @@ all    		 :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:
 clean  		 :; dapp clean
 test   		 :; ./test.sh match=$(match) optimizer=1
 test-dev   	 :; ./test.sh match=$(match) optimizer=0
```
naszam, line 4: same as above

iamchrissmith: you mean adding quotes?

naszam:
```suggestion
test   		 :; ./test.sh match="$(match)" optimizer=1
test-dev   	 :; ./test.sh match="$(match)" optimizer=0
```

src/DssDirectDepositHub.sol
```diff
@@ -298,7 +288,7 @@ contract DssDirectDepositHub {
         if (assetBalance > daiDebt) {
             fees = assetBalance - daiDebt;
```
iamchrissmith, line 289: Use `unchecked { ... }` here and in other places where we don't need safemath

brianmcmichael: Can't comment on the line since it's not updated, but in the Hub exec function, can also do:

`toUnwind = Art - lineWad;` as unchecked I think.

src/plans/D3MAaveDaiPlan.sol
brianmcmichael: I definitely don't want to hinder progress toward the later solidity version on behalf of dss-exec-lib. We can probably use the rpc tests for this for now until we have a way to manage different solc versions.

src/plans/D3MAaveDaiPlan.sol
```diff
@@ -154,19 +145,16 @@ contract D3MAaveDaiPlan is ID3MPlan {
         // Do inverse calculation of interestStrategy
         uint256 variableRateSlope1 = interestStrategy.variableRateSlope1();
         uint256 targetUtil;
-        if (targetInterestRate > _add(base, variableRateSlope1)) {
+        if (targetInterestRate > base + variableRateSlope1) {
             // Excess interest rate
             uint256 r = targetInterestRate - base - variableRateSlope1;
```
iamchrissmith, line 150: unchecked

src/plans/D3MAaveDaiPlan.sol
```diff
@@ -154,19 +145,16 @@ contract D3MAaveDaiPlan is ID3MPlan {
...
-                            interestStrategy.OPTIMAL_UTILIZATION_RATE()
-                        );
+            targetUtil = _rdiv(
+                            _rmul(interestStrategy.EXCESS_UTILIZATION_RATE(), r),
+                            interestStrategy.variableRateSlope2()
+                        ) + interestStrategy.OPTIMAL_UTILIZATION_RATE();
         } else {
             // Optimal interest rate
-            targetUtil = _rdiv(_rmul(_sub(targetInterestRate, base), interestStrategy.OPTIMAL_UTILIZATION_RATE()), variableRateSlope1);
+            targetUtil = _rdiv(_rmul((targetInterestRate - base), interestStrategy.OPTIMAL_UTILIZATION_RATE()), variableRateSlope1);
```
iamchrissmith, line 157: refactor to be multiline

Makefile
```diff
@@ -1,6 +1,6 @@
 all    		 :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.6.12 build
 clean  		 :; dapp clean
-test   		 :; ./test.sh match=$(match) optimizer=1
-test-dev   	 :; ./test.sh match=$(match) optimizer=0
-test-forge 	 :; ./test-forge.sh match=$(match) block=$(block) match-test=$(match-test) match-contract=$(match-contract)
+test   		 :; ./test.sh match="$(match)" optimizer=1
+test-dev   	 :; ./test.sh match="$(match)" optimizer=0
+test-forge 	 :; ./test-forge.sh match="$(match)" block="$(block)" match-test="$(match-test)" match-contract="$(match-contract)"
 deploy 		 :; make && dapp create DssDirectDepositAaveDai 0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F 0x4449524543542d4141564556322d444149000000000000000000000000000000 0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9 0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5
```
gbalabasquer, line 6: Not the correct PR but I guess as we are changing other stuff here we could remove this `deploy` command which is old.

src/DssDirectDepositHub.sol
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity 0.6.12;
+pragma solidity >=0.6.12;
```
gbalabasquer, line 17: Keeping the pragma this way is actually a bug. As compiling with solc 0.6.x will make the safe math is not applied and then the code has a complete different behaviour. Not sure if 0.7.0 is already implementing it but to be honest I'd just do:
```
pragma solidity ^0.8.13;
```

src/DssDirectDepositHub.sol
```diff
@@ -530,7 +519,7 @@ contract DssDirectDepositHub {
         address vow_ = vow;
         uint256 wad = vat.gem(ilk, address(pool));
         require(wad < 2 ** 255, "DssDirectDepositHub/overflow");
-        vat.suck(vow_, vow_, _mul(wad, RAY)); // This needs to be done to make sure we can deduct sin[vow] and vice in the next call
+        vat.suck(vow_, vow_, (wad * RAY)); // This needs to be done to make sure we can deduct sin[vow] and vice in the next call
```
gbalabasquer, line 522:
```suggestion
        vat.suck(vow_, vow_, wad * RAY); // This needs to be done to make sure we can deduct sin[vow] and vice in the next call
```

src/DssDirectDepositHub.sol
```diff
@@ -406,7 +396,6 @@ contract DssDirectDepositHub {
                     currentAssets
                 );
             } else {
-                // All the subtractions are safe as otherwise toUnwind is > 0
                 uint256 toWind = targetAssets - currentAssets;
                 toWind = _min(toWind, lineWad - Art);
                 toWind = _min(toWind, (Line - debt) / RAY);
```
gbalabasquer, line 401: As Brian pointed out above, this can be inside `unchecked`

src/plans/D3MAaveDaiPlan.sol
```diff
@@ -175,17 +163,14 @@ contract D3MAaveDaiPlan is ID3MPlan {
...
-                TokenLike(dai).balanceOf(adai),
-                totalDebt
-            );
+        uint256 totalPoolSize = TokenLike(dai).balanceOf(adai) + totalDebt;
 
         uint256 targetTotalPoolSize = _calculateTargetSupply(targetInterestRate, totalDebt);
         if (targetTotalPoolSize >= totalPoolSize) {
             // Increase debt (or same)
-            return _add(currentAssets, targetTotalPoolSize - totalPoolSize);
+            return currentAssets + (targetTotalPoolSize - totalPoolSize);
```
gbalabasquer, line 173: in theory we could separate the math here to make the subtraction unchecked but doesn't seem to be worth it.

src/DssDirectDepositHub.sol
hexonaut: Yes strongly agree with @gbalabasquer here.

Makefile
```diff
@@ -1,6 +1,6 @@
 all    		 :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.6.12 build
```
talbaneth, line 1:
```suggestion
all    		 :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.8.13 build
```

src/plans/D3MAaveDaiPlan.sol
```diff
@@ -154,19 +145,16 @@ contract D3MAaveDaiPlan is ID3MPlan {
...
-                            _rdiv(
-                                _rmul(interestStrategy.EXCESS_UTILIZATION_RATE(), r),
-                                interestStrategy.variableRateSlope2()
-                            ),
-                            interestStrategy.OPTIMAL_UTILIZATION_RATE()
-                        );
+            targetUtil = _rdiv(
+                            _rmul(interestStrategy.EXCESS_UTILIZATION_RATE(), r),
+                            interestStrategy.variableRateSlope2()
+                        ) + interestStrategy.OPTIMAL_UTILIZATION_RATE();
```
talbaneth, line 154: Indentation changed here to 3 spaces instead of 4.
Also the closing bracket of the `_rdiv` is not completely aligned to it (missing one space before the closing bracket).

src/plans/D3MAaveDaiPlan.sol
talbaneth: Also consider removing the brackets around `targetInterestRate - base`

src/plans/D3MAaveDaiPlan.sol
```diff
@@ -175,17 +163,14 @@ contract D3MAaveDaiPlan is ID3MPlan {
...
         if (targetInterestRate == 0) return 0;  // De-activated
 
-        uint256 totalDebt = _add(stableDebt.totalSupply(), variableDebt.totalSupply());
+        uint256 totalDebt = stableDebt.totalSupply() + variableDebt.totalSupply();
 
-        uint256 totalPoolSize = _add(
-                TokenLike(dai).balanceOf(adai),
-                totalDebt
-            );
+        uint256 totalPoolSize = TokenLike(dai).balanceOf(adai) + totalDebt;
```
talbaneth, line 168:
```suggestion
        uint256 totalPoolSize = dai.balanceOf(adai) + totalDebt;
```

src/plans/D3MAaveDaiPlan.sol
gbalabasquer: I hadn't seen this comment before, but as I mentioned in mine, we must not leave a pragma that allows to compile the contract in a broken state and that's what means leaving the `>=0.6.12` one here.
`dss-exec-lib` will need to figure it out how to include this dependency, might be better just a mock or worst case maintain a branch of this repo for it.

Makefile
iamchrissmith: i'd prefer to keep this out of this PR.  probably belongs in v2

shell.nix
```diff
@@ -1,11 +1,11 @@
 { url
   , dappPkgs ? (
     import (fetchTarball "https://github.com/makerdao/makerpkgs/tarball/master") {}
-  ).dappPkgsVersions.hevm-0_43_1
+  ).dappPkgsVersions.master-20220406
```
gbalabasquer, line 4:
```suggestion
  ).dappPkgsVersions.master-20220524
```

shell.nix
```diff
@@ -1,11 +1,11 @@
...
 }: with dappPkgs;
 
 mkShell {
-  DAPP_SOLC = solc-static-versions.solc_0_6_12 + "/bin/solc-0.6.12";
+  DAPP_SOLC = solc-static-versions.solc_0_8_13 + "/bin/solc-0.8.13";
```
gbalabasquer, line 8:
```suggestion
  DAPP_SOLC = solc-static-versions.solc_0_8_14 + "/bin/solc-0.8.14";
```

src/DssDirectDepositHub.sol
```diff
@@ -223,7 +213,7 @@ contract DssDirectDepositHub {
     }
 
     /**
-        @notice remove nope on daiJoin to kill hub
+        @notice remove hope on daiJoin to kill hub
```
gbalabasquer, line 216: We might want to revert this change and evaluate this function in another PR. I think this is useless. Removing the `vat` access to the `daiJoin` should not be really needed and we should avoid adding non useful functions.

src/DssDirectDepositHub.sol
```diff
@@ -378,16 +371,22 @@ contract DssDirectDepositHub {
...
             uint256 toUnwind;
             if (Art > lineWad) {
-                toUnwind = Art - lineWad;
+                unchecked {
+                    toUnwind = Art - lineWad;
+                }
             }
             if (debt > Line) {
-                toUnwind = _max(toUnwind, _divup(debt - Line, RAY));
+                unchecked {
```
gbalabasquer, line 379: Was wondering if the `unchecked` might modify the behaviour of `_divup`, but no, nothing to worry about:
"The setting only affects the statements that are syntactically inside the block. Functions called from within an unchecked block do not inherit the property."
https://docs.soliditylang.org/en/v0.8.14/control-structures.html

src/DssDirectDepositHub.sol
```diff
@@ -306,7 +299,7 @@ contract DssDirectDepositHub {
         require(amount <= 2 ** 255, "DssDirectDepositHub/overflow");
```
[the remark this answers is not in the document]
gbalabasquer: Just to clarify, the net result is the same for solc 0.8.x as will revert anyway, I'm proposing it reverts in the line that actually checks the overflow than later.
If we would like to replicate the exact same behaviour than with solc < 0.8.0, we should complicate the code and is not worth it for that extra wei.
I know for this specific function we are probably moving to the extra value on `min` but that can be addressed later in another PR from this better meaning of the `require`.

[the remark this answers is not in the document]
gbalabasquer: On `uncull` we should also change the format of the require.

test.sh
```diff
@@ -10,16 +10,16 @@ do
...
+            *)
     esac
 done
 
 export DAPP_BUILD_OPTIMIZE="$OPTIMIZER"
 export DAPP_BUILD_OPTIMIZE_RUNS=200
 
 if [[ -z "$1" ]]; then
-  dapp --use solc:0.6.12 test --rpc-url="$ETH_RPC_URL" -v
+  dapp --use solc:0.8.13 test --rpc-url="$ETH_RPC_URL" -v
```
gbalabasquer, line 22:
```suggestion
  dapp --use solc:0.8.14 test --rpc-url="$ETH_RPC_URL" -v
```

test.sh
```diff
@@ -10,16 +10,16 @@ do
...
 else
-  dapp --use solc:0.6.12 test --rpc-url="$ETH_RPC_URL" --match "$MATCH" -vv
+  dapp --use solc:0.8.13 test --rpc-url="$ETH_RPC_URL" --match "$MATCH" -vv
```
gbalabasquer, line 24:
```suggestion
  dapp --use solc:0.8.14 test --rpc-url="$ETH_RPC_URL" --match "$MATCH" -vv
```

src/DssDirectDepositHub.sol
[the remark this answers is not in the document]
gbalabasquer: I think `require(amount <= type(int256).max, "DssDirectDepositHub/overflow");` is the best format.

gbalabasquer:
@iamchrissmith you might want to merge https://github.com/makerdao/dss-direct-deposit/pull/49 to v2 and then v2 to this before merging this PR.

src/DssDirectDepositHub.sol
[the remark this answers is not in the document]
iamchrissmith: done https://github.com/makerdao/dss-direct-deposit/pull/47/commits/70e46b4ede0ec98ac6949e2d780c3ed9b5cb7b17

gbalabasquer (approved):
LGTM

src/DssDirectDepositHub.sol
```diff
@@ -442,10 +441,12 @@ contract DssDirectDepositHub {
                     currentAssets
                 );
             } else {
-                // All the subtractions are safe as otherwise toUnwind is > 0
```
naszam, line 445: why it was removed?

iamchrissmith: i think i removed it before `unchecked` was added so safemath was in place.  adding it back

gbalabasquer (approved):
Renewing approval.

src/plans/D3MPlanBase.t.sol
```diff
@@ -88,7 +88,7 @@ contract D3MPlanBaseTest is DSTest {
             address(bytes20(uint160(uint256(keccak256("hevm cheat code")))))
         );
 
-        dai = DaiLike(123);
+        dai = DaiLike(address(123));
```
naszam, line 91:
```suggestion
        dai = DaiLike(address(0x123));
```

src/pools/D3MAaveDaiPool.sol
```diff
@@ -158,7 +151,7 @@ contract D3MAaveDaiPool is ID3MPool {
         // Verify the correct amount of adai shows up
         uint256 interestIndex = pool.getReserveNormalizedIncome(address(asset));
         uint256 scaledAmount = _rdiv(wad, interestIndex);
-        return adai.scaledBalanceOf(address(this)) >= _add(scaledPrev, scaledAmount);
+        return adai.scaledBalanceOf(address(this)) >= scaledPrev + scaledAmount;
```
naszam, line 154:
```suggestion
        return adai.scaledBalanceOf(address(this)) >= (scaledPrev + scaledAmount);
```

src/plans/D3MPlanBase.t.sol
```diff
@@ -88,7 +88,7 @@ contract D3MPlanBaseTest is DSTest {
             address(bytes20(uint160(uint256(keccak256("hevm cheat code")))))
         );
 
-        dai = DaiLike(address(123));
+        dai = DaiLike(address(0x123));
```
gbalabasquer, line 91: It's pretty common we do `address(123)` in general, it just means something random.
I mean not against to the change, but don't think it is a fix at all.

src/pools/D3MAaveDaiPool.sol
```diff
@@ -151,7 +151,7 @@ contract D3MAaveDaiPool is ID3MPool {
         // Verify the correct amount of adai shows up
         uint256 interestIndex = pool.getReserveNormalizedIncome(address(asset));
         uint256 scaledAmount = _rdiv(wad, interestIndex);
-        return adai.scaledBalanceOf(address(this)) >= scaledPrev + scaledAmount;
+        return adai.scaledBalanceOf(address(this)) >= (scaledPrev + scaledAmount);
```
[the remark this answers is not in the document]
gbalabasquer: The sum happens before the comparison.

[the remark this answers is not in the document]
gbalabasquer: I think we even removed the parenthesis from same places where wasn't necessary (not every place though).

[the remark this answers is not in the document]
iamchrissmith: i understood @naszam 's comment to suggest that the previous method could have been evaluated as:
1. is scaledBalanceOf >= scaledPrev -> 0 or 1 (bool)
2. add scaledAmount to 0 or 1

So this ensures we do the math first.  0.8 did change some order of operations things, but even if this wasn't a bug, i think the `(...)` is a little clearer and doesn't cost anything (right?)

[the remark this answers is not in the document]
naszam: I left a suggestion for consistency with the previous function `_add` that was returning the sum, I would consider it a nit, I should have clarified that.

src/plans/D3MPlanBase.t.sol
naszam: yeah, same as above, was meant to be a nit

src/pools/D3MAaveDaiPool.sol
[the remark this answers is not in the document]
naszam: I don't have a strong opinion on this, if without parenthesis is cleaner I'm ok with that.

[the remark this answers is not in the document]
iamchrissmith: @gbalabasquer also do not feel strongly here.  You pick with `(...)` or without

src/plans/D3MPlanBase.t.sol
iamchrissmith: i'd revert then for consistency with our other dummy addresses

iamchrissmith: e76057a

naszam (approved):
LGTM

PR #49, opened 2022-05-11
Forge ci

iamchrissmith:
Use Forge Test for CI


iamchrissmith:
@gbalabasquer I moved this change to a separate PR since you discussed with Vamsi a fix to 0.8.13 and Dapp that will hopefully help make #47 work

.github/workflows/tests.yaml
```diff
@@ -1,25 +1,18 @@
...
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
 
-    - name: Use maker and dapp cachix
-      uses: cachix/cachix-action@v10
-      with:
-        name: maker
-        extraPullNames: dapp
+      - name: Install deps
```
gbalabasquer, line 14: Hmm I think this name should be removed and then just attach the `run: forge install` to the "name: Install Foundry" block

gbalabasquer, line 14: Hmm I think this name should be removed and then just attach the `run: forge install` to the "name: Install Foundry" block

test-forge-blocks.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
gbalabasquer, line 1: Hmmm was there a discussion to add this test? I think it will be too heavy and not sure if worth it (in fact failed for this run).
I'd leave this PR to the very minimum which is switching from dapptools to foundry.

iamchrissmith: moved those pieces to a new PR so we can discuss if we want/how we should do it (https://github.com/makerdao/dss-direct-deposit/pull/53)

.github/workflows/tests.yaml
iamchrissmith: I tried this and GH actions errors with a `uses` and `run` in the same block.  Changed the names for better clarity

gbalabasquer: Sounds good!

test-forge-blocks.sh
gbalabasquer: perfect, thanks!

gbalabasquer (approved):
I've updated the rpc key and running the tests again. If they pass it looks good to me for merging.

.github/workflows/tests.yaml
```diff
@@ -1,25 +1,18 @@
...
-      with:
-        submodules: recursive
+      - uses: actions/checkout@v3
 
-    - name: Install nix 2.3.6
-      uses: cachix/install-nix-action@v13
-      with:
-        install_url: https://releases.nixos.org/nix/nix-2.3.6/install
-        nix_path: nixpkgs=channel:nixos-unstable
+      - name: Define Foundry
```
naszam, line 9:
```suggestion
      - name: Install Foundry
```

.github/workflows/tests.yaml
```diff
@@ -1,25 +1,18 @@
...
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
 
-    - name: Use maker and dapp cachix
-      uses: cachix/cachix-action@v10
-      with:
-        name: maker
-        extraPullNames: dapp
+      - name: Install Foundry
```
naszam, line 14:
```suggestion
      - name: Install Dependencies
```

naszam (approved):
Notes: left two minor step name suggestions

LGTM

.github/workflows/tests.yaml
gbalabasquer: Is it actually installing anything in this block or is it defining where to get foundry/forge?

gbalabasquer: Excepting I'm not understanding how it works I feel the naming Chris changed is more appropriate.
Actually your suggestion is getting back to what it was.

naszam: I've checked it against here: https://github.com/marketplace/actions/foundry-toolchain

naszam: I think the github action is installing foundry

gbalabasquer: Hmm if that is the case, then is right, why are we also adding `forge install`?

gbalabasquer: Ahh ok got it. Then yes, these suggested names (similar to the original ones) make more sense.
Sorry about the confusion.

naszam: no worries! got confused as well initially.

gbalabasquer:
@iamchrissmith I think after applying the changes Naz made (basically rolling back) we are ready for merging.

naszam (approved):
LGTMerge

gbalabasquer (approved):
LGTM. The failure is due actual tests failing so it's ok to merge.

PR #46, opened 2022-05-06
D3M version 2.0

iamchrissmith:
Update D3M from v1 (used for Aave) to v2.  V2 includes D3M code for Aave and Compound. For more details about the new system see: https://github.com/makerdao/dss-direct-deposit/blob/7c706e254a466632da033a9c21bfafee7a353433/README.md

PR #51, opened 2022-05-17
Compound

talbaneth:


hexonaut (commented):
Finished review on the pool. Will do the plan next.

src/pools/D3MCompoundDaiPool.t.sol
```diff
@@ -0,0 +1,227 @@
...
+        pool.file("fail", address(123));
+    }
+
+    function test_deposit_calls_cdai_deposit() public {
+        assertEq(cDai.balanceOfUnderlying(address(pool)), 0);
+        uint256 poolBefore = dai.balanceOf(address(pool));
+
+        pool.deposit(1 * WAD);
+
+        assertEqApproxBPS(cDai.balanceOfUnderlying(address(pool)), 1 * WAD, 1);
```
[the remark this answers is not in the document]
talbaneth: Done.
For the underlying dai comparisons I moved to use 10^-9 instead of 1 bps (0.3 dai for 300m deposit).

src/pools/D3MCompoundDaiPool.t.sol
```diff
@@ -0,0 +1,227 @@
...
+
+    function test_withdraw_calls_cdai_withdraw() public {
+        pool.deposit(1 * WAD);
+
+        assertEqApproxBPS(cDai.balanceOfUnderlying(address(pool)), 1 * WAD, 1);
+        uint256 before = dai.balanceOf(address(this));
+
+        pool.withdraw(1 * WAD);
+
+        assertEqApproxBPS(cDai.balanceOfUnderlying(address(pool)), 0, 1);
```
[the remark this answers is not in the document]
talbaneth: Done (also in other places).

src/pools/D3MCompoundDaiPool.t.sol
```diff
@@ -0,0 +1,227 @@
...
+import { DaiLike, TokenLike }    from "../tests/interfaces/interfaces.sol";
+import { D3MCompoundDaiPool }    from "./D3MCompoundDaiPool.sol";
+
+interface CErc20Like {
+    function balanceOf(address owner) external view returns (uint256);
+    function comptroller()            external view returns (uint256);
+    function balanceOfUnderlying(address owner) external returns (uint256);
+}
+
+interface CompltrollerLike {
```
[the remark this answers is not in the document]
talbaneth: Done.

src/plans/D3MCompoundDaiPlan.t.sol
```diff
@@ -0,0 +1,304 @@
...
+        assertEq(plan.getTargetAssets(0), 0);
+    }
+
+    function test_getTargetAssets_current_rate() public {
+        cDai.accrueInterest();
+        uint256 borrowRatePerBlock = cDai.borrowRatePerBlock();
+        plan.file("barb", borrowRatePerBlock);
+
+        uint256 targetAssets = plan.getTargetAssets(0);
+        assertEqAbsolute(0, targetAssets, WAD);
```
hexonaut, line 272: Actual and expected should be flipped.

hexonaut (changes requested):
Previous issues addressed. Finished full review.

src/plans/D3MCompoundDaiPlan.sol
```diff
@@ -0,0 +1,199 @@
...
+// util <= kink:
+//         targetInterestRate = util * multiplierPerBlock / WAD + baseRatePerBlock
+// (2)     util = (targetInterestRate - baseRatePerBlock) * WAD / multiplierPerBlock
+//
+// https://github.com/compound-finance/compound-protocol/blob/master/contracts/BaseJumpRateModelV2.sol#L79
+//
+//         util = borrows * WAD / (cash + borrows - reserves);
+// (3)     cash + borrows - reserves = borrows * WAD / util
+
+pragma solidity >=0.6.12;
```
naszam, line 45:
```suggestion
pragma solidity ^0.6.12;
```

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.6.12;
```
naszam, line 17:
```suggestion
pragma solidity ^0.6.12;
```

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+interface CanLike {
+    function hope(address) external;
+    function nope(address) external;
+}
+
+interface D3mHubLike {
+    function vat() external view returns (address);
+}
+
+interface CErc20Like is TokenLike {
```
naszam, line 36: Consider adding a comment with the link to the full interface

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+    modifier auth {
+        require(wards[msg.sender] == 1, "D3MCompoundDaiPool/not-authorized");
+        _;
+    }
+
+    // --- Events ---
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+    event File(bytes32 indexed what, address data);
+    event Collect(address indexed king, address indexed comp, uint256 amt);
```
naszam, line 81:
```suggestion
    event Collect(address indexed king, address indexed gift, uint256 amt);
```

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+
+    function hope(address hub) external override auth {
+        CanLike(D3mHubLike(hub).vat()).hope(hub);
+    }
+
+    function nope(address hub) external override auth {
+        CanLike(D3mHubLike(hub).vat()).nope(hub);
+    }
+
+    function deposit(uint256 amt) external override auth {
```
naszam, line 135: Consider including NatSpec comments

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+        uint256 prev = cDai.balanceOf(address(this));
+        require(cDai.mint(amt) == 0, "D3MCompoundDaiPool/mint-failure");
+        require(
+            cDai.balanceOf(address(this)) ==
+            _add(prev, _wdiv(amt, cDai.exchangeRateStored())), "D3MCompoundDaiPool/incorrect-cdai-credit"
```
naszam, line 140: Consider adding a comment with a rational for using `exchangeRateStored`

src/plans/D3MCompoundDaiPlan.sol
```diff
@@ -0,0 +1,199 @@
...
+        uint256 jumpMultiplierPerBlock = rateModel.jumpMultiplierPerBlock();
+
+        uint256 normalRate = _add(_wmul(kink, multiplierPerBlock), baseRatePerBlock);
+
+        uint256 targetUtil;
+        if (targetInterestRate > normalRate) {
+            if (jumpMultiplierPerBlock == 0) return 0; // illegal rate, max is normal rate for this case
+            targetUtil = _add(kink, _wdiv(targetInterestRate - normalRate, jumpMultiplierPerBlock));             // (1)
+        } else if (targetInterestRate > baseRatePerBlock) {
+            targetUtil = _wdiv(targetInterestRate - baseRatePerBlock, multiplierPerBlock);                       // (2)
```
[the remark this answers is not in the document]
hexonaut: Ok calculation makes sense. Still there is a discrepancy with how Aave is handled in that bad values of `bar` will cause this function to revert in Aave vs in here it will just return 0. We should unify this design. I think using your conversion of illegal values to return 0 target supply instead of reverting is a good idea.

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    function accrueIfNeeded() override external {
+        require(cDai.accrueInterest() == 0, "D3MCompoundDaiPool/accrueInterest-failure");
+    }
+
+    // Does not accrue interest (as opposed to cToken's balanceOfUnderlying() which is not a view function).
+    function assetBalance() public view override returns (uint256) {
+        (uint256 error, uint256 cTokenBalance,, uint256 exchangeRate) = cDai.getAccountSnapshot(address(this));
+        return (error == 0) ? _wmul(cTokenBalance, exchangeRate) : 0;
```
talbaneth, line 164: Should revert on an error, not return 0.

src/plans/D3MCompoundDaiPlan.sol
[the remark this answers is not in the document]
talbaneth: Resolving this as implemented on AAVE V2 branch.

src/plans/D3MCompoundDaiPlan.sol
talbaneth: Done.

src/pools/D3MCompoundDaiPool.sol
```diff
@@ -0,0 +1,201 @@
...
+contract D3MCompoundDaiPool is ID3MPool {
+
+    mapping (address => uint256) public wards;
+    address public king; // Who gets the rewards
+
+    ComptrollerLike public immutable comptroller;
+    TokenLike       public immutable dai;
+    CErc20Like      public immutable cDai;
+
+    // --- Auth ---
```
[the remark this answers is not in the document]
talbaneth: Done.

src/pools/D3MCompoundDaiPool.sol
talbaneth: Done.

src/pools/D3MCompoundDaiPool.sol
talbaneth: Done.

src/plans/D3MCompoundDaiPlan.sol
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    // TODO: this function seems unneeded, remove once it's removed from the interface and AAVE
+    // targetSupply = cash + borrows - reserves
+    function calculateTargetSupply(uint256 targetInterestRate) external view returns (uint256) {
+        return _calculateTargetSupply(targetInterestRate, cDai.totalBorrows());
+    }
+
+    function _calculateTargetSupply(uint256 targetInterestRate, uint256 borrows) internal view returns (uint256) {
```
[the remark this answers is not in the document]
talbaneth: Resolving as this is now implemented in `v2` branch.

src/pools/D3MCompoundDaiPool.sol
talbaneth: I think this is redundant as there is already natspec documentation on the interface where this function is declared.
Let me know if anyone thinks differently.

src/pools/D3MCompoundDaiPool.sol
talbaneth: Done.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+        return _wdiv(borrows, targetUtil);                                                      // (3)
+    }
+
+    function active() public view override returns (bool) {
+        if (barb == 0) return false;
+        return CErc20Like(cDai).interestRateModel() == address(tack);
+    }
+
+    function disable() external override {
+        require(wards[msg.sender] == 1 || !active(), "D3MCompoundPlan/not-authorized");
```
kmbarry1, line 183: What's the reasoning behind allowing anyone to disable if the there's an interest rate model mismatch? I don't really see an issue with it right now, but just wondering if it's actually _necessary_ for some reason?

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+        if (what == "barb") {
+            barb = data;
+        } else revert("D3MCompoundPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
+    function file(bytes32 what, address data) external auth {
+        if (what == "tack") tack = InterestRateModelLike(data);
+        else revert("D3MCompoundPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
```
kmbarry1, line 126: Why have a `file` that allows setting an arbitrary `tack` when it needs to be the cDai interest rate model anyway? At a minimum it seems this `file` should `require` the new `tack` be equal to `CErc20Like(cDai_).interestRateModel()`.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+
+import "./ID3MPlan.sol";
+
+// https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/CErc20.sol#L14
+interface CErc20Like {
+    function totalBorrows()           external view returns (uint256);
+    function totalReserves()          external view returns (uint256);
+    function interestRateModel()      external view returns (address);
+    function getCash()                external view returns (uint256);
+    function underlying()             external view returns (address);
```
brianmcmichael, line 56: `underlying()` Not used.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+    function kink()                   external view returns (uint256);
+    function multiplierPerBlock()     external view returns (uint256);
+    function jumpMultiplierPerBlock() external view returns (uint256);
+}
+
+contract D3MCompoundPlan is ID3MPlan {
+
+    mapping (address => uint256) public wards;
+    InterestRateModelLike        public tack;
+    uint256                      public barb; // target Interest Rate Per Block [wad] (0)
```
brianmcmichael, line 71: What does this number look like? If governance wants to target, say, a 2.75% annual rate, what's the formula for calculating that? We might need to document this formula.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+        uint256 targetUtil;
+        if (targetInterestRate > normalRate) {
+            if (jumpMultiplierPerBlock == 0) return 0; // illegal rate, max is normal rate for this case
+            targetUtil = kink + _wdiv(targetInterestRate - normalRate, jumpMultiplierPerBlock); // (1)
+        } else if (targetInterestRate > baseRatePerBlock) {
+            targetUtil = _wdiv(targetInterestRate - baseRatePerBlock, multiplierPerBlock);      // (2)
+        } else {
+            // if (target == base) => (borrows == 0) => supply does not matter
+            // if (target  < base) => illegal rate
+            return 0;
```
kmbarry1, line 169: Hmmmm...not sure about this behavior. I'd naively think the way to interpret `targetInterestRate <= baseRatePerBlock` would be "inject liquidity up to available debt ceiling to make the rate as close to the target as possible". Instead, we'll unwind in such a state, maximizing the interest rate instead of minimizing it. Ofc I guess one can just argue that governance or what is setting the target interest rate should just be taking this into account, but IMO it's a counterintuitive edge case. E.g. I'd expect setting the target interest rate to zero would be equivalent to "inject max allowed liquidity".

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+        wards[usr] = 1;
+        emit Rely(usr);
+    }
+    function deny(address usr) external auth {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
+
+    function file(bytes32 what, uint256 data) external auth {
+        if (what == "barb") {
```
brianmcmichael, line 117: Add appropriate boundary here. I think that means that `data < WAD` but Tal will double-check.

src/plans/D3MCompoundPlan.sol
brianmcmichael: Ideally it would be nice to have an `external pure` function in this Plan such that a developer can pass basis points as an argument and get an appropriate value to be used for setting `barb`

src/plans/D3MCompoundPlan.sol
talbaneth: Just discussed it with the others. The idea was that if the rate model on Aave/Compound changes anyone can permanently disable the module, so that even ithe model on Aave/Compound returns to the original one it will still be disabled.
Motivation was to not support this kind of back and forth model changing.

src/plans/D3MCompoundPlan.sol
talbaneth: The new `tack` should be a specific one (i.e a parameter and not fetched from `cDai`) since we only want to support a model that was reviewed by our side.

We can add the `require` you suggested but in any case if there is a mismatch then `active()` check will fail and will cause unwinding (and not supporting the new unknown model).
A secondary reasoning here is that we do usually try to make the `file` functions as flat/minimal as possible.

Let me know if you still think we need the `require` (which will fail the spell but not cause unwinding).

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+
+// https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/BaseJumpRateModelV2.sol#L10
+interface InterestRateModelLike {
+    function baseRatePerBlock()       external view returns (uint256);
+    function kink()                   external view returns (uint256);
+    function multiplierPerBlock()     external view returns (uint256);
+    function jumpMultiplierPerBlock() external view returns (uint256);
+}
+
+contract D3MCompoundPlan is ID3MPlan {
```
kmbarry1, line 67: Has there been any research done on whether there's any risk of reentrancy on the CToken being used to manipulate the results? Unfotunately the reentrancy guard variable on a CToken is not public...and I think the only functions marked non-reentrant are not state-modifying (meaning we couldn't have `view` functions here if we called one of them). There may be no issue anyway, but was curious if anyone besides me had thought about it.

talbaneth: I'm not aware of such research (probably not for AAVE as well). We should look into that.
Can you think of what an attacker has to gain other than causing unintended (and possibly multiple) winds/unwinds?

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,187 @@
...
+// https://github.com/compound-finance/compound-protocol/blob/master/contracts/BaseJumpRateModelV2.sol#L79
+//
+//         util = borrows * WAD / (cash + borrows - reserves);
+// (3)     cash + borrows - reserves = borrows * WAD / util
+
+pragma solidity ^0.8.14;
+
+import "./ID3MPlan.sol";
+
+// https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/CErc20.sol#L14
```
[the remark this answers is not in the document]
talbaneth: Yeah as mentioned I didn't want to point to that guy's private repo, as this was finally merged yesterday we can update the link.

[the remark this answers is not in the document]
talbaneth: btw for reference about their recent cDai deployment - https://compound.finance/governance/proposals/104

src/plans/D3MCompoundPlan.sol
talbaneth: We can use this value - https://github.com/compound-finance/compound-protocol/blob/a3214f67b73310d547e00fc578e8355911c9d376/contracts/CTokenInterfaces.sol#L31

src/pools/D3MCompoundPool.sol
```diff
@@ -0,0 +1,199 @@
...
+
+    function collect() external {
+        require(king != address(0), "D3MCompoundPool/king-not-set");
+
+        address[] memory holders = new address[](1);
+        holders[0] = address(this);
+        address[] memory cTokens = new address[](1);
+        cTokens[0] = address(cDai);
+
+        comptroller.claimComp(holders, cTokens, false, true);
```
talbaneth, line 189: Consider fetching the comptroller from cDai everytime here instead of having it in storage.

src/plans/D3MCompoundPlan.sol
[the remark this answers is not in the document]
naszam: sharing here if you'd like to follow a similar approach for compound:
https://github.com/makerdao/dss-direct-deposit/pull/46#discussion_r892845549

src/pools/D3MCompoundDaiPool.sol
naszam: yeah, maybe the comment could be restricted to the compound case, following a similar approach as per aave (see https://github.com/makerdao/dss-direct-deposit/blob/v2/src/pools/D3MAavePool.sol#L147-L148)

naszam: same for `withdraw`

src/plans/D3MCompoundPlan.sol
kmbarry1: I need to do deeper research into know what the worst-case consequences are. One thing to consider is that other contracts in the future might call the plan's `view` functions and use the results in various ways--it's hard to predict what the consequences could be for such contracts of manipulation.

src/plans/D3MCompoundPlan.sol
kmbarry1: Hmmmm I guess not having a `require` gives us some flexibility in case we've reviewed a plan before Compound has officially switched to it and want to go ahead and `file` it. Arguably we should have a whitelist of plans so that we can smoothly migrate with no disruptions, similar to what is done for upgradeable ERC20 gems. So...I argree no `require` needed, but now you need to talk me out of having a whitelist instead of a unique value.

src/plans/D3MCompoundPlan.sol
kmbarry1: Okay but why don't we want to support re-enabling if the other protocol returns to a trusted plan? Just to force governance to re-evaluate? Also do we care if no one bothers to call `disable` in such a situation?

talbaneth: Yeah this has been discussed in the V2 reviews, the idea is to force governance to re-evaluate.
We indeed need someone to call `disable` for not supporting returning to an old supported model. It was decided we prefer that then the alternative of making `active` a non view function.

src/plans/D3MCompoundPlan.sol
talbaneth: Yeah this was also suggested and turned down since we didn't want to support protocols to return to an old supported version without us unwinding. The downside of unwinding and waiting for governance was deemed as not too bad. Personally I think we might want to support the whitelist. If we want to re-open the discussion it should be done in the `v2` PR.

src/plans/D3MCompoundPlan.sol
talbaneth: Yeah it makes sense. As this is also relevant for AAVE let's continue the discussion here - https://github.com/makerdao/dss-direct-deposit/pull/46/files#r893886318.

src/plans/D3MCompoundPlan.sol
talbaneth: done in `36eb5e109a`

src/plans/D3MCompoundPlan.sol
talbaneth: Done in 0cdf45a.

src/pools/D3MCompoundPool.sol
talbaneth: I eventually opted out of this and even make `comp` immutable.
Since we later have `comp.transfer(king, amt)` I feel it's more secured to have both `comptroller` and `comp` immutable (otherwise a malicious comptroller can allow sending out cdai, even though that line should send to the pause proxy).

@iamchrissmith fyi (I think in aave the situation with `rewardsClaimer` is less risky as there is no additional transfer step).

src/pools/D3MCompoundDaiPool.sol
talbaneth: Yeah I know in Aave we have additional comments for both `withdraw` and `deposit`.
I think I'd rather keep things clean in the implementation main part, and only comment on non trivial stuff (like the comment we do have now inside `deposit`).
Don't think the existence of comments should be exactly the same through Aave/Compound pools/plans.

naszam: I think it's a good practise, in general, to have comments for external integrators and ease code archaeology

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,198 @@
...
+        emit File(what, data);
+    }
+    function file(bytes32 what, address data) external auth {
+        if (what == "tack") tack = InterestRateModelLike(data);
+        else if (what == "delegate") delegate = data;
+        else revert("D3MCompoundPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    function getTargetAssets(uint256 currentAssets) external override view returns (uint256) {
```
gbalabasquer, line 138: nit: can set the same order than aave plan where first is `_calculateTargetSupply` than `getTargetAssets` ?

gbalabasquer: There are also minor differences in `getTargetAssets` that could be exact same code for both, like spacing, order, use of parenthesis or not. If we could standardize it for at least these two is much better when doing diffs.
Maybe the Compound one is better and we should just change in Aave one.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    function _calculateTargetSupply(uint256 targetInterestRate, uint256 borrows) internal view returns (uint256) {
+        uint256 kink                   = tack.kink();
+        uint256 multiplierPerBlock     = tack.multiplierPerBlock();
+        uint256 baseRatePerBlock       = tack.baseRatePerBlock();
+        uint256 jumpMultiplierPerBlock = tack.jumpMultiplierPerBlock();
+
+        uint256 normalRate = _wmul(kink, multiplierPerBlock) + baseRatePerBlock;
```
gbalabasquer, line 168: I think it'd be better called as `maxNormalRate`

talbaneth: This is actually a Compound term:
https://github.com/compound-finance/compound-protocol/blob/master/contracts/BaseJumpRateModelV2.sol#L102
Tried to stick with their vocabulary.

gbalabasquer: I see this is how is called in Compound, but at least confused me for a while...

gbalabasquer: Sorry my last comment was without reloading and hadn't seen yours. Yeah I saw the same after. I guess it is fine to keep the same naming, maybe just a comment to clarify would be fine.

src/plans/D3MCompoundPlan.sol
gbalabasquer: Hmm I think we defined for this repo to put all the internals before. In fact the Aave plan doesn't follow it correctly as was defined for the Hub neither. But I definitely wouldn't take what was in DssVest as a defined standard globally.
So I'd we come at least for some standard within this repo and in this case that Compound and Aave are very similar try to keep the code differences the minimum as possible. For both `_calculateTargetSupply` should be moved right after internal math functions.

gbalabasquer: I'll take care of it in `v2` branch so then you can pull mirror both the best is possible.

gbalabasquer: I was actually confused, in Hub we have first the administration external functions than the internals which is actually the same than the actual Aave plan. I'd say then, for now we stick to that, so nothing to be done in `v2`.

naszam: @talbaneth, I confirm this was discussed on `v2` and the general consensus was following a different code layout pattern for d3m, diverging from `dss-vest` and the canonical in solc docs, grouping internal functions before targeted external ones (aka Gonza's standard) IIRC.

src/pools/D3MCompoundPool.sol
```diff
@@ -0,0 +1,200 @@
...
+        comptroller = ComptrollerLike(cDai.comptroller());
+        comp        = TokenLike(comptroller.getCompAddress());
+
+        require(address(comp) != address(0), "D3MCompoundPool/invalid-comp");
+
+        dai.approve(cDai_, type(uint256).max);
+
+        hub = hub_;
+        vat = VatLike(D3mHubLike(hub).vat());
+        vat.hope(hub);
```
gbalabasquer, line 86: nit:
```suggestion
        vat.hope(hub_);
```
It is more gas efficient, and same way than done in aave.

gbalabasquer, line 86: nit
```suggestion
        vat = VatLike(D3mHubLike(hub_).vat());
        vat.hope(hub_);
```
It is more gas efficient, and same way than done in aave.

src/pools/D3MCompoundPool.sol
```diff
@@ -0,0 +1,200 @@
...
+        require(error == 0, "D3MCompoundPool/getAccountSnapshot-failure");
+        return _wmul(cTokenBalance, exchangeRate);
+    }
+
+    function maxDeposit() external pure override returns (uint256) {
+        return type(uint256).max;
+    }
+
+    function maxWithdraw() external view override returns (uint256) {
+        return _min(cDai.getCash(), assetBalance());
```
gbalabasquer, line 177: I was wondering if the reserves amount could influence the max amount to withdraw but from what I followed in the code, whatever DAI is available as cDai balance can be withdrawn. So all should be fine.
Just wanted to make sure this was also checked from your side and we can close this comment.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,198 @@
...
+    event File(bytes32 indexed what, address data);
+
+    constructor(address cDai_) {
+        address rateModel_ = CErc20Like(cDai_).interestRateModel();
+        address delegate_  = CErc20Like(cDai_).implementation();
+
+        require(rateModel_ != address(0), "D3MCompoundPlan/invalid-rateModel");
+        require(delegate_  != address(0), "D3MCompoundPlan/invalid-delegate");
+
+        cDai     = CErc20Like(cDai_);
```
naszam, line 92: Consider moving this line at the top to avoid redundant cast, following compound pool pattern.

README.md
```diff
@@ -35,3 +35,14 @@ Below are the configurable parameters for the Aave DAI D3M:
...
 
 Any stkAave that is accured can be permissionlessly collected into the pause proxy by calling `collect()`.
+
+## Compound D3M
+
+### Configuration
+
+Below are the configurable parameters for the Compound DAI D3M:
+
+- `tau` [seconds] - The expiry for when bad debt is sent to the vow debt queue. This must be set during initialization to enforce a deadline for when this module is considered to be in a failure mode where no more liquidity is available in the pool to unwind. Unwinding can still occur after this period elapses.
```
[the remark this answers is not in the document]
talbaneth: Yeah, that's better imo. Done in `e9bbcbb2ffeb32`

src/plans/D3MCompoundPlan.sol
talbaneth: Added a comment at `c9167a684`

src/pools/D3MCompoundPool.sol
talbaneth: Gas seems to be the same, changed it for alignment with aave in `8ce3449`.

src/plans/D3MCompoundPlan.sol
talbaneth: Done in `cdd0a82254`

src/pools/D3MCompoundPool.sol
talbaneth: Yeah that's right.
In any given time:
total amount of withdrawal rights = cash + borrows - reserves.

So if you withdraw the entire balance (`cash`) it just means that the reserves are borrowed out.
But still after you do that the amount of borrows will be the total amount of withdrawal rights + reserves.

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,200 @@
...
+            targetUtil = kink + _wdiv(targetInterestRate - normalRate, jumpMultiplierPerBlock); // (1)
+        } else if (targetInterestRate > baseRatePerBlock) {
+            targetUtil = _wdiv(targetInterestRate - baseRatePerBlock, multiplierPerBlock);      // (2)
+        } else {
+            // if (target == base) => (borrows == 0) => supply does not matter
+            // if (target  < base) => illegal rate
+            return 0;
+        }
+
+        if (targetUtil > WAD) return 0; // illegal rate (unacheivable utilization)
```
gbalabasquer, line 160: As discussed this should go away.

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,342 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity ^0.8.14;
+
+import {DSSTest} from "dss-test/DSSTest.sol";
```
gbalabasquer, line 20: Placing it here as I can't comment in the dependency itself. But this was downgraded from `v2`.

gbalabasquer: And if we merge will downgrade it there.

src/pools/D3MCompoundPool.sol
gbalabasquer: Ah yes? so there isn't a SLOAD in the constructor. Good to know.

src/plans/D3MCompoundPlan.sol
gbalabasquer: Ok sounds good, we didn't discuss that but we can do in these days. If Compund way is better we can change the Aave one.

src/tests/pools/D3MCompoundPool.t.sol
```diff
@@ -0,0 +1,282 @@
...
+    function test_deposit_calls_cdai_deposit() public {
+        assertEq(cDai.balanceOfUnderlying(address(pool)), 0);
+        assertEq(cDai.balanceOf(address(pool)), 0);
+        uint256 poolBefore = dai.balanceOf(address(pool));
+
+        pool.file("hub", address(this));
+        pool.deposit(1 * WAD);
+
+        _assertEqApprox(cDai.balanceOfUnderlying(address(pool)), 1 * WAD);
+        assertGt(cDai.balanceOf(address(pool)), 0);
```
gbalabasquer, line 133: This is pointless checking the one below.

talbaneth: It does make sure that `_wdiv(1 * WAD, cDai.exchangeRateStored())` does not result in 0 and would go unnotice in this test.
Are you convinced it's worthless?

src/tests/pools/D3MCompoundPool.t.sol
```diff
@@ -0,0 +1,282 @@
...
+        uint256 before = pool.assetBalance();
+        pool.file("hub", address(this));
+        pool.deposit(1 * WAD);
+        _assertEqApprox(pool.assetBalance() - before, 1 * WAD);
+    }
+
+    function test_maxWithdraw_gets_available_assets() public {
+        pool.file("hub", address(this));
+        pool.deposit(1 * WAD);
+        _assertEqApprox(pool.assetBalance(), pool.maxWithdraw());
```
gbalabasquer, line 276:
```suggestion
        assertEq(pool.assetBalance(), pool.maxWithdraw());
```

src/tests/plans/D3MCompoundPlan.t.sol
talbaneth: Fixed in `907c41f4`

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,346 @@
...
+        return _calculateTargetSupply(targetInterestRate, cDai.totalBorrows());
+    }
+}
+
+contract D3MCompoundPlanTest is D3MPlanBaseTest {
+    CErc20Like             cDai;
+    InterestRateModelLike  model;
+    D3MCompoundPlanWrapper plan;
+
+    function _add(uint256 x, uint256 y) internal pure returns (uint256 z) {
```
gbalabasquer, line 59: All safe math needs to go away.

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        cTokens[0] = address(cEth);
+
+        ComptrollerLike(cDai.comptroller()).enterMarkets(cTokens);
+    }
+
+    // --- Math ---
+    function _min(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = x <= y ? x : y;
+    }
+    function _add(uint256 x, uint256 y) internal pure returns (uint256 z) {
```
gbalabasquer, line 161: All safe math needs to go away.

src/tests/pools/D3MCompoundPool.t.sol
```diff
@@ -0,0 +1,282 @@
...
+    ComptrollerLike comptroller;
+    TokenLike       comp;
+    LensLike        lens;
+
+    function _mul(uint256 x, uint256 y) public pure returns (uint256 z) {
+        require(y == 0 || (z = x * y) / y == x, "ds-math-mul-overflow");
+    }
+
+    function _wdiv(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = _mul(x, WAD) / y;
```
gbalabasquer, line 52:
```suggestion
        z = x * WAD / y;
```

src/plans/D3MCompoundPlan.sol
```diff
@@ -0,0 +1,199 @@
+// SPDX-FileCopyrightText: © 2021-2022 Dai Foundation <www.daifoundation.org>
```
naszam, line 1: Following recent guidance clarification from Dai Foundation on discord, it emerged that only the year of file creation would need to be noted.
```suggestion
// SPDX-FileCopyrightText: © 2021 Dai Foundation <www.daifoundation.org>
```

src/pools/D3MCompoundPool.sol
```diff
@@ -0,0 +1,200 @@
+// SPDX-FileCopyrightText: © 2021-2022 Dai Foundation <www.daifoundation.org>
```
naszam, line 1: Same as above.
```suggestion
// SPDX-FileCopyrightText: © 2021 Dai Foundation <www.daifoundation.org>
```

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,346 @@
...
+        plan.file("delegate", address(1));
+
+        assertEq(plan.delegate(), address(1));
+    }
+
+    function test_cannot_file_unknown_address_param() public {
+        assertRevert(d3mTestPlan, abi.encodeWithSignature("file(bytes32,address)", bytes32("bad"), address(1)), "D3MCompoundPlan/file-unrecognized-param");
+    }
+
+    function test_cannot_file_without_auth() public {
```
gbalabasquer, line 182: Missing this test for `file(bytes32,address)`

gbalabasquer: And btw I'd have the `file(bytes32,uint256)` grouped with the other ones.

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,346 @@
...
+        cDai.accrueInterest();
+        uint256 borrowRatePerBlock = cDai.borrowRatePerBlock();
+        plan.file("barb", borrowRatePerBlock);
+
+        uint256 targetAssets = plan.getTargetAssets(0);
+        assertEqAbsolute(targetAssets, 0, WAD);
+    }
+
+    function test_rate_model_changed_not_active() public {
+        // Simulate Compound changing the rate model in the pool
```
gbalabasquer, line 302:
```suggestion
        // Simulate Compound changing the rate model in the pool
        assertTrue(address(plan.tack()) == cDai.interestRateModel());
        assertTrue(plan.active());
```

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,346 @@
...
+        // Simulate Compound changing the cDai implementation
+        assertTrue(address(plan.delegate()) == cDai.implementation());
+        plan.file("delegate", address(456));
+
+        assertTrue(address(plan.delegate()) != cDai.implementation());
+        assertTrue(plan.active() == false);
+    }
+
+    function test_barb_zero_not_active() public {
+        assertEq(plan.barb(), 0);
```
gbalabasquer, line 319:
```suggestion
        assertTrue(plan.active());
        assertEq(plan.barb(), 0);
```

src/tests/plans/D3MCompoundPlan.t.sol
gbalabasquer: Actually we also need to set barb so it doesn't fail due that.

src/plans/D3MCompoundPlan.sol
gbalabasquer: For `Hub` and  `Aave` is arguable as the logic itself comes from the older version which was done last year.

src/pools/D3MCompoundPool.sol
gbalabasquer: Same.

src/tests/pools/D3MCompoundPool.t.sol
gbalabasquer: Yeah I thought that might be the reason why was added, however I do not see a practical way than `_wdiv(1 * WAD, cDai.exchangeRateStored()` can be absolute 0.
Anyway this was a comment that honestly I should not have even done hehe.

src/pools/D3MCompoundPool.sol
naszam: Updated.

src/plans/D3MCompoundPlan.sol
naszam: There are still the old copyright comments though, should those be removed in favour of the new format?

naszam: Following latest guidance updates, the old copyright format should be replaced with the new one.

src/pools/D3MCompoundPool.sol
```diff
@@ -0,0 +1,200 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2021-2022 Dai Foundation
```
naszam, line 3: Same as above.
```suggestion
```

src/plans/D3MCompoundPlan.sol
naszam: Left suggestions below.

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        assertTrue(amountSupplied > 0);
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(d3mCompoundPool));
+        assertEqRounding(ink, amountSupplied);
+        assertEqRounding(art, amountSupplied);
+
+        assertEq(vat.gem(ilk, address(d3mCompoundPool)), 0);
+        assertEq(vat.dai(address(d3mHub)), 0);
+    }
+
+    function test_target_increase() public {
```
gbalabasquer, line 315: Maybe we should cache the `amountSupplied` before raising the rate to prove after doing it will be lower.

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+
+        // Assert that indeed current utilization > 100%
+        assertGt(rateModel.utilizationRate(cDai.getCash(), cDai.totalBorrows(), cDai.totalReserves()), WAD);
+
+        // File the current rate, make sure it is supported (does not cause unwind) although utilization is over 100%
+        d3mCompoundPlan.file("barb", cDai.borrowRatePerBlock());
+        d3mHub.exec(ilk);
+        _assertEqApprox(initialSupply, cDai.balanceOfUnderlying(address(d3mCompoundPool)), WAD / 100);
+
+        // File a the maximum supported rate
```
gbalabasquer, line 379:
```suggestion
        // File the maximum supported rate
```

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        end.cage();
+
+        if (originalSin + part * RAY >= originalDai) {
+            assertEq(vat.sin(vow), originalSin + part * RAY - originalDai);
+            assertEq(vat.dai(vow), 0);
+        } else {
+            assertEq(vat.dai(vow), originalDai - originalSin - part * RAY);
+            assertEq(vat.sin(vow), 0);
+        }
+
```
gbalabasquer, line 845: We could add an `assertRevert` here to show that `end.cage(ilk)` won't work until `uncull(ilk)` happens

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        d3mHub.reap(ilk); // Clear out fees at the start
+
+        (uint256 pink, uint256 part) = vat.urns(ilk, address(d3mCompoundPool));
+        uint256 gemBefore = vat.gem(ilk, address(d3mCompoundPool));
+        uint256 viceBefore = vat.vice();
+        uint256 sinBefore = vat.sin(vow);
+        uint256 vowDaiBefore = vat.dai(vow);
+        uint256 cdaiDaiBalanceBefore = dai.balanceOf(address(cDai));
+        uint256 poolAdaiBalanceBefore = cDai.balanceOf(address(d3mCompoundPool));
```
gbalabasquer, line 1109: Should be `poolCdaiBalanceBefore`

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        uint256 initialRate = _setRelBorrowTarget(5000);
+        d3mHub.reap(ilk); // Clear out fees at the start
+
+        (uint256 pink, uint256 part) = vat.urns(ilk, address(d3mCompoundPool));
+        uint256 gemBefore = vat.gem(ilk, address(d3mCompoundPool));
+        uint256 viceBefore = vat.vice();
+        uint256 sinBefore = vat.sin(vow);
+        uint256 vowDaiBefore = vat.dai(vow);
+        uint256 cdaiDaiBalanceBefore = dai.balanceOf(address(cDai));
+        uint256 poolAdaiBalanceBefore = cDai.balanceOf(address(d3mCompoundPool));
```
gbalabasquer, line 1163: Should be `poolCdaiBalanceBefore`

src/plans/D3MCompoundPlan.sol
talbaneth: This is not a trivial calculation and adding it on-chain should be discussed separately.
If we do add it a proper place can also be dss-exec-lib.
This is not a blocker as we can start with a manual calculation or a rates translation chart.

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1225 @@
+// SPDX-FileCopyrightText: © 2021-2022 Dai Foundation <www.daifoundation.org>
```
naszam, line 1: Same as above for tests, consider keeping file creation year (for compound it should be 2022) and removing old copyright comment (as per Dai Foundation copyright guidance).

src/tests/plans/D3MCompoundPlan.t.sol
talbaneth: Handled in `c6cf814`

src/tests/integration/D3MCompound.t.sol
talbaneth: Handled in `c6cf814`

src/tests/pools/D3MCompoundPool.t.sol
```diff
@@ -0,0 +1,282 @@
...
+
+contract D3MCompoundPoolTest is D3MPoolBaseTest {
+
+    CErc20Like      cDai;
+    D3MCompoundPool pool;
+    ComptrollerLike comptroller;
+    TokenLike       comp;
+    LensLike        lens;
+
+    function _mul(uint256 x, uint256 y) public pure returns (uint256 z) {
```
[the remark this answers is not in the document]
talbaneth: Handled in `c6cf814`

src/tests/plans/D3MCompoundPlan.t.sol
```diff
@@ -0,0 +1,346 @@
...
+        // Simulate Compound changing the rate model in the pool
+        plan.file("tack", address(456));
+
+        assertTrue(address(plan.tack()) != cDai.interestRateModel());
+        assertTrue(plan.active() == false);
+    }
+
+    function test_delegate_changed_not_active() public {
+        // Simulate Compound changing the cDai implementation
+        assertTrue(address(plan.delegate()) == cDai.implementation());
```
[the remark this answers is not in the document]
talbaneth: Handled in `eba23f7018`

src/tests/plans/D3MCompoundPlan.t.sol
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
```diff
@@ -0,0 +1,1234 @@
...
+        // Accumulate a bunch of interest
+        hevm.roll(block.number + 180 * 5760);
+
+        // Someone else borrows almost all the liquidity
+        assertEq(cDai.borrow(dai.balanceOf(address(cDai)) - 100 * WAD), 0);
+
+        // Reap the partial fees
+        uint256 vowDai = vat.dai(vow);
+        d3mHub.reap(ilk);
+        assertEq(vat.dai(vow), vowDai + 100 * RAD);
```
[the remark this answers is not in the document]
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
talbaneth: Handled in `eba23f7018`

src/tests/integration/D3MCompound.t.sol
naszam: Sounds good.

src/tests/pools/D3MCompoundPool.t.sol
gbalabasquer: Let's leave it as it is.

gbalabasquer (approved):
I think this PR is in good shape to be squashed and merged to `v2`.
I still think there might be some potential improvements we could to make more uniformed the Compund and Aave plans and pools but that can be addressed in a separate PR as it might exceed Compound code.

src/plans/D3MCompoundPlan.sol
```diff
@@ -37,163 +37,162 @@
...
+        if (targetInterestRate == 0) return 0; // De-activated
+
+        uint256 borrows = cDai.totalBorrows();
+        uint256 targetTotalPoolSize = _calculateTargetSupply(targetInterestRate, borrows);
+        uint256 totalPoolSize = cDai.getCash() + borrows - cDai.totalReserves();
 
-        return _wdiv(borrows, targetUtil);                                                                       // (3)
+        if (targetTotalPoolSize >= totalPoolSize) {
+            // Increase debt (or same)
+            return currentAssets + targetTotalPoolSize - totalPoolSize;
```
hexonaut, line 173: For additional protection against overflow it's better to subtract the delta first before performing the addition.

hexonaut (changes requested):
One minor change, but rest LGTM.

src/plans/D3MCompoundPlan.sol
talbaneth: Done in `75d0a7a4f1`

hexonaut (approved):

src/plans/D3MCompoundPlan.sol
talbaneth: Did some changes in V2 and here to better align code style.
Not everything is fully aligned but I think diffing is easier now.

PR #53, opened 2022-05-24
Multiple block tests

iamchrissmith:
Adds a multiple block test runner for the CI.  This looks at the last 50 blocks and runs our test suite (using Forge) against that to see if the tests pass with different block states.

If we like this, one possible better solution would be to define a range and then randomly select blocks in that range so we get better "state coverage"


test-forge-blocks.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+BLOCK=$(cast block-number)
```
gbalabasquer, line 4: I'm wondering if we could just combine both scripts in one and just have two workflows. The regular one will run from latest to latest block.

.github/workflows/block-tests.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - name: Install deps
+        run: forge install
+
+      - name: Give Permission
+        run: chmod u+x ./test-forge-blocks.sh
```
naszam, line 18: wondering if this is needed.

PR #56, opened 2022-06-02
Add rectify to repay bad debt along with tests

hexonaut:
If a loss occurs the hub will attempt to re-enforce the position up to the debt ceiling. This will allow governance to take action and move the bad debt to the surplus buffer. In the event of a hack the debt ceilings protect us and we would not want to call this. In the event of an expected loss (such as loan default in Maple) it may make sense for governance to repay the debt and reach the desired target exposure again.

### DIFF
--- a/src/D3MHub.t.sol
+++ b/src/D3MHub.t.sol
@@ -430,7 +430,7 @@ contract D3MHubTest is DSTest {
         assertEq(art, 50 * WAD);
 
         // This should only fill another 10 because the debt ceiling
-        directDepositHub.exec(ilk);
+        d3mHub.exec(ilk);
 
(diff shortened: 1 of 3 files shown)


gbalabasquer:
Need to look it in detail, but we were discussing the option of removing `quit` in favor of potentially doing it through a spell (to reduce code). This might enter in the same category and we might prefer to use the spell (potentially a dss-exec-lib function) to take care of it.

hexonaut:
I can understand with `quit()` as that is more a thing that should only be called when something goes wrong. `rectify` on the other hand should be expected to be called during normal operation as we start venturing into territory where defaults are more likely.

gbalabasquer:
We decided to go with this functionality outside the `Hub` for now. Closing this PR.

PR #59, opened 2022-06-03
Add D3MOracle

gbalabasquer:
This PR is Sam's idea as a workaround to fix the `uncull` issue.
Basically it blocks `pip.read()` in `End.cage(ilk)` until makes sure the `ilk` is `uncull`ed.

This has some caveats, D3M ilks should use this special Oracle in the `spotter` instead of a common `DSValue`. More importantly it decouples the functionality of `read` from `peek`. The normal behavior is having `read` to revert when `peek` returns `false` as the second parameter. In this case the response diverges as we use `peek` for the `spotter.poke(ilk)` operations (normal case where we do not want to put a price of zero even if the `ilk` is `cull`ed) and `read` for `end.cage(ilk)` where we want to halt execution if the `ilk` is culled.

It is not the most elegant solution but probably the only realistic way if we do not want to modify `End`.

Third option is removing the `uncull` function and accept the collateral is lost if this case happens.

Last one is accept the race condition, which IMO is the worst option and I would be opposed to it.


gbalabasquer:
We are going the path of modifying the `End`. Then closing this PR.

gbalabasquer:
Reopening as we are back to this idea. Need to merge v2, solve conflicts and some minor changes. So for now a WIP.

gbalabasquer:
This PR should be ready for review.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        val = WAD;
+        ok = vat.live() == 1 || HubLike(hub).culled(ilk) == 0;
+    }
+
+    /**
+        @notice Return value
+        @dev vat must be live or ilk must be unculled in hub.
+        @return val always 1 WAD value
+    */
+    function read() external view returns (uint256 val) {
```
kmbarry1, line 112: I worry a bit about unintended consequences here, even though its behavior is only deviant after emergency shutdown. I guess the "limited impact" solution would be to have the End `peek` and check `ok` itself.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        else revert("D3MOracle/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    /**
+        @notice Return value and status of the oracle
+        @return val always 1 WAD
+        @return ok true if vat is live or ilk is not culled
+    */
+    function peek() public view returns (uint256 val, bool ok) {
```
kmbarry1, line 102: Should we have a `peep` too just for ABI compatibility? Probably not, but want to cover all bases.

kmbarry1:
A couple philosophical questions to consider, but probably not blockers. Code basically LGTM.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        val = WAD;
+        ok = vat.live() == 1 || HubLike(hub).culled(ilk) == 0;
```
talbaneth, line 104: Subjective (and we might want to ignore at this stage) but an alternative of having the logic for `ok` here is calling a new view function in the hub (`hub.endable(ilk)`?) and having the logic there.

The conditions for `ok` are really entwined with the hub states, so this will support updating the hub without updating this oracle. Will also save us from having `vat` here and add some clarity for this condition through the function name (though for that we can just add a comment here).

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+        @notice Return value
+        @dev vat must be live or ilk must be unculled in hub.
```
talbaneth, line 109:
```suggestion
        @dev vat must be live or ilk must not be culled  in hub.
```
`unculled` is a bit confusing here (as it's not only when `unculled` was called but also when `culled` was not called).
Alternatively can just say `peek must return true for ok` as the conditions for that are already detalied in `peek`.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+    */
+    function deny(address usr) external auth {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
+
+    /**
+        @notice update vow or end addresses.
+        @dev msg.sender must be authorized.
+        @param what name of what we are updating bytes32("vow"|"end")
```
talbaneth, line 86:
```suggestion
        @param what name of what we are updating.
```

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        @notice Return value
+        @dev vat must be live or ilk must be unculled in hub.
+        @return val always 1 WAD value
+    */
+    function read() external view returns (uint256 val) {
+        bool ok;
+        (val, ok) = peek();
+        require(ok, "D3MOracle/ilk-is-culled");
+    }
+}
```
talbaneth, line 117: We are removing the ds-value's `void()` option (to set `ok` to 0). I don't see a problem with that (price is constant, no liquidations, in ES we don't want gov intervention anyhow), but pointing it in case someone sees a problem.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        @dev msg.sender must be authorized.
+        @param usr address to be de-authorized
+    */
+    function deny(address usr) external auth {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
+
+    /**
+        @notice update vow or end addresses.
```
talbaneth, line 84:
```suggestion
        @notice update an address.
```

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+        @notice Return value
+        @dev vat must be live or ilk must be unculled in hub.
+        @return val always 1 WAD value
+    */
+    function read() external view returns (uint256 val) {
+        bool ok;
+        (val, ok) = peek();
+        require(ok, "D3MOracle/ilk-is-culled");
```
talbaneth, line 115:
```suggestion
        require(ok, "D3MOracle/not-ok");
```
`ok` can currently also be false if vat is not live.

src/D3MOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+
+contract D3MOracle {
+    // --- Auth ---
+    /**
+        @notice Maps address that have permission in the Pool.
+        @dev 1 = allowed, 0 = no permission
+        @return authorization 1 or 0
+    */
+    mapping (address => uint256) public wards;
+    address public hub;
```
talbaneth, line 36: Can define this as `HubLike` (as we are doing that for `vat`).

src/D3MOracle.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        assertTrue(ok);
+        vat.cage();
+        (value, ok) = oracle.peek();
+        assertEq(value, WAD);
+        assertTrue(!ok);
+        hub.uncull();
+        (value, ok) = oracle.peek();
+        assertEq(value, WAD);
+        assertTrue(ok);
+    }
```
talbaneth, line 132: Missing a `peek` check for when `vat` is not caged but `hub` is culled (should be !ok).

talbaneth (commented):
See small comments/suggestions.

src/D3MOracle.sol
gbalabasquer: Hmmm I guess this is a workaround, and we trust the oracle implements we want to stop execution.
I don't think using `peek` it actually marks a difference, as at the end if peek is also bad implemented and returns true when it shouldn't we will have the same effect.

src/D3MOracle.sol
gbalabasquer: Not really, this replaces a `DSValue` which doesn't have `peep`. There is no concept of OSM for this as we do not have for example on PSMs.

src/D3MOracle.sol
gbalabasquer: We still need `vat` for the `file` function so I'd say we keep as it is.
If we update the `hub` we still need to update the address in the oracles so the only difference would be we need to make sure to not break anything of the structure and being everything in the same repo makes it quite resistant to that kind of errors.

src/D3MOracle.sol
gbalabasquer: Done!

src/D3MOracle.sol
gbalabasquer: Yeah it should be fine, it is kind of a "hardcoded" feed price.

src/D3MOracle.sol
gbalabasquer: Not really, needs to be vat is not live `and` ilk not culled.
I wanted to be very explicit on the error message to realize what is missing. In case we get to this one the solution is to `uncull`.

src/D3MOracle.sol
gbalabasquer: Not really opposed to do it but here my reasoning on why I left this way:
- vat is immutable and used twice, both times used calling a method.
- hub is used once and I liked to keep `file` without castings.

Anyway if you feel looks better I'll proceed.

src/D3MOracle.t.sol
gbalabasquer: Not sure about this one, this is exactly the case I wanted to avoid reporting `false`. That's why originally I wasn't including `vat.live` in the check and then doing `peek` behavior different than `read`.

If the system is in normal operation, and the collateral is culled we do not gain anything on start reporting an invalid price to the spotter. That will make the vat.spot for that ilk is set to 0. I don't see that can be a problem as once the `ilk` is culled in the `Hub` is irreversible for normal operation, but feeling we better keep the 1 value in vat.spot.

hexonaut (approved):
LGTM

src/D3MOracle.sol
talbaneth: What I meant was that ok is only false if vat is not live and ilk is culled, so "ilk-is-culled" is a bit missleading (or missing information). If we want to be detailed, maybe change to "ilk-culled-in-shutdown"?

src/D3MOracle.sol
talbaneth: Ok. Maybe only add a comment of how this relates to the End?
Just to remove the required knowledge for understanding why the condition was added.

src/D3MOracle.sol
talbaneth: Yeah I fill the uniformity in the variable definiton looks a bit better and casting in `file` is something we do a lot.
So in favor on changing, but not critical at all.

src/D3MOracle.t.sol
talbaneth: Right.

src/D3MOracle.sol
gbalabasquer: Changed!

src/D3MOracle.sol
gbalabasquer: Ok good, I've proceed giving uniformity but changing the `vat` to an `address`. Hope that looks good as well.

src/D3MOracle.sol
godsflaw: I suppose the DSValue is so different than other oracle implementations that preserving the interface is already a moot topic, but not having `void()`, not using storage variables for `has` and `val`, just means we are absolutely going to need to establish a new `class` pattern for these in the exec lib.  I bring this up just in case someone else can think of an edge case where this is a blocker.  I can't.

godsflaw (approved):
LGTM

talbaneth (approved):
LGTM

src/D3MOracle.sol
gbalabasquer: Well this is an oracle that doesn't take more configuration than the `hub`, so there is not `poke` or `void`, just reading functions.

gbalabasquer: A collateral already can have a `DSValue` an `Osm`, a `LPOsm` (with the different variants) and could even have directly a `Median`. The only real shared interface between all of them are `read` and `peek`.

PR #63, opened 2022-06-06
remove Pool.active and clean up

iamchrissmith:
remove the pool.active function

add conditions to the plan.active function

clean up spacing

### DIFF
--- a/src/plans/D3MAavePlan.sol
+++ b/src/plans/D3MAavePlan.sol
@@ -119,7 +119,7 @@ contract D3MAavePlan is ID3MPlan {
         else revert("D3MAavePlan/file-unrecognized-param");
         emit File(what, data);
     }
-    
+
     function file(bytes32 what, uint256 data) external auth {
         if (what == "bar") bar = data;
         else revert("D3MAavePlan/file-unrecognized-param");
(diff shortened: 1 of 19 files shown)


talbaneth:
This might not be ready for review, but can I ask what is the idea with the new `wild` getter?

iamchrissmith:
> This might not be ready for review, but can I ask what is the idea with the new `wild` getter?

It came up in the red team review as a way we can permissionlessly cage a pool/plan that is misconfigured or misbehaving.  

Tbh I am not liking all these different states for pools, plans, d3ms etc. But thought we could review from this PR and try to combine them

iamchrissmith:
per @talbaneth in compound review: should make sure we document active and disable relationship

src/plans/D3MAavePlan.sol
```diff
@@ -200,6 +207,19 @@ contract D3MAavePlan is ID3MPlan {
         return strategy == address(tack);
     }
 
+    function pause() external override {
+        paused_ = true;
+    }
```
iamchrissmith, line 212: needs to be `auth` and add tests

src/plans/D3MAavePlan.sol
```diff
@@ -194,8 +193,13 @@ contract D3MAavePlan is ID3MPlan {
 
     function active() public view override returns (bool) {
         if (bar == 0) return false;
-        (,,,,,,,,,, address strategy,) = pool.getReserveData(address(dai));
-        return strategy == address(tack);
+        (,,,,,,, address adai_, address stableDebt_, address variableDebt_, address strategy,) = pool.getReserveData(address(dai));
+        return (
+                    strategy == address(tack) &&
```
naszam, line 198: nit: consider aligning `&`

iamchrissmith: ae618a4

PR #65, opened 2022-06-08
rename getter functions and clear shadow warnings

iamchrissmith:
following convention here: https://github.com/makerdao/dss-vest/blob/master/src/DssVest.sol#L82-L112

rename the getter functions and clear variable shadow warnings

### DIFF
--- a/src/D3MHub.sol
+++ b/src/D3MHub.sol
@@ -224,7 +224,7 @@ contract D3MHub {
     }
 
     // --- Deposit controls ---
-    function _wind(bytes32 ilk, ID3MPool pool, uint256 amount) internal {
+    function _wind(bytes32 ilk, ID3MPool pool_, uint256 amount) internal {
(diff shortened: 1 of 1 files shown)


src/D3MHub.sol
```diff
@@ -224,7 +224,7 @@ contract D3MHub {
     }
 
     // --- Deposit controls ---
-    function _wind(bytes32 ilk, ID3MPool pool, uint256 amount) internal {
+    function _wind(bytes32 ilk, ID3MPool pool_, uint256 amount) internal {
```
naszam: Consider putting the underscore as prefix to follow `dss-vest` pattern.
```suggestion
    function _wind(bytes32 ilk, ID3MPool _pool, uint256 amount) internal {
```

naszam (approved):
LGTM

gbalabasquer (approved):
LGTM. Just not sure what is the convention to put the _ before or after the variable name. Is it the same a variable passed by param than a memory variable?

PR #66, opened 2022-06-09
Implement MIP21 as a D3M

hexonaut:
This PR re-implements [MIP21](https://github.com/makerdao/MIP21-RWA-Example) as a D3M. Currently this is a PoC, but could be fairly easily made real.

One fundamental difference is that MIP21 sets the interest rate whereas the D3M model leaves the interest rate to the external party/protocol. However, that being said in the case of 6S they are actually setting the interest rate anyways and the interest set by the protocol is a facade. So really having the external party/protocol set the interest rate is what's being done all around anyways.

I'll go over how each piece translates:

### RwaUrn

`D3MHub` automatically covers the minting/burning of ERC20 DAI. I didn't add a `gem` lock, but this is trivial to include if desired.

### RwaOutputConduit  and RwaInputConduit

These contracts are merged into `D3MMIP21Plan`. Most functions are the same `RwaInputConduit.push()` has been renamed to `pull()` to avoid function name collision. DAI funds are pulled and pushed to the `D3MMIP21Pool` which I'm aliasing as a generic `escrow`.

Order of operations for winding up the position:

1. Governance sets `cap` parameter to desired debt ceiling. Ex) 20m DAI
2. Operator sets the `to` address in the plan.
3. Anyone with MKR calls `push()` to send the `cap` amount to the `to` address and decrease `cap` to indicate all repayments should immediately proceed to repay the debt.

Order of operations for winding down the position:

1. When receiver of the loan wants to repay they simply send DAI to the plan address.
2. Anyone with MKR calls `pull()` which forwards the DAI to the pool. Keepers will automatically detect that DAI liquidity is available at that point and withdraw/repay the debt.

### RwaLiquidationOracle

Liquidation is implemented in a nearly identical way inside `D3MHub` when you call `cage(ilk)`.


src/plans/D3MMIP21Plan.sol
```diff
@@ -0,0 +1,156 @@
...
+    function hope(address usr) external auth {
+        can[usr] = 1;
+        emit Hope(usr);
+    }
+    function nope(address usr) external auth {
+        can[usr] = 0;
+        emit Nope(usr);
+    }
+
+    function file(bytes32 what, uint256 data) external auth {
```
godsflaw, line 96: Might actually want this to be the `operator` modifier.  This would allow governance to set an `ilk.line` or `vat.Line` to limit exposure, and the `operator` to set `cap` to the maximum debt they want to pull (or us to inject) at any given time.

src/plans/D3MMIP21Plan.sol
```diff
@@ -0,0 +1,156 @@
...
+        to = who;
+        emit Pick(who);
+    }
+    function push() external {
+        require(to != address(0), "D3MMIP21Plan/to-not-set");
+        require(gov.balanceOf(msg.sender) > 0, "D3MMIP21Plan/no-gov");
+        uint256 balance = dai.balanceOf(address(escrow));
+        uint256 amtToPush = _min(balance, cap);
+        unchecked {
+            cap -= amtToPush;
```
godsflaw, line 144: I think `cap` should just be a limit that determines if `balance` is still under `cap`.

src/plans/D3MMIP21Plan.sol
```diff
@@ -0,0 +1,156 @@
...
+    function disable() external override {
+        require(
+            wards[msg.sender] == 1 ||
+            !active()
+        , "D3MMIP21Plan/not-authorized");
+        cap = 0;
+        emit Disable();
+    }
+
+    // --- Routing ---
```
godsflaw, line 132: I think these routing functions should be on the pool, since that maintains balances rather than wind/unwind logic.  We may even just leave this as conduits, and add onto the conduits the functions the plan or pool need to function.

src/plans/D3MMIP21Plan.sol
```diff
@@ -0,0 +1,156 @@
...
+    event Nope(address indexed usr);
+    event Kiss(address indexed who);
+    event Diss(address indexed who);
+    event Pick(address indexed who);
+    event Push(address indexed to, uint256 wad);
+
+    constructor(address dai_, address gov_, address escrow_) {
+        dai = TokenLike(dai_);
+        gov = TokenLike(gov_);
+        escrow = escrow_;
```
godsflaw, line 56: What is escrow in this context?  `to` should be the broker-dealer where the funds are sent.

godsflaw (commented):
not a full review, just some comments.  I think we need to push this a little further, and talk about rates and protocol income to fully understand if this is a superset of MIP21.

src/plans/D3MMIP21Plan.sol
hexonaut: Going to explain my mental model of Plan vs Pool with some pretty pictures. TLDR is pool should be dumb and stateless. Plan should have all logic / state.

src/plans/D3MMIP21Plan.sol
hexonaut: Escrow = D3MMIP21Pool.

iamchrissmith:
how are fees accrued here? is it just that they send more dai than borrowed based on the agreement without an on-chain component?

dbkbali (commented):
A few points for clarification:
- most of the MIP21s - I have reviewed so far have off-chain valuations for the underlying collateral - with the ERC20 token acting as a dummy placeholder, and not representing on-chain the true value of the underlying collateral - wondering how you would incorporate this in the plan?
- If D3M is about adjusting the amount of DAI invested to adjust the underlying demand/supply curve for DAI liquidity - not sure how this achieves that as it doesn't target a DAI borrowing rate but appears to target a level of underlying tokens that we want to hold?
- what is the incentive to ensure any DAI minted into the target protocol that this generates underlying yield for MDAO - specifically in the proposed Maple case - the lpTokens that MD would hold will only generate yield if there are actual loans funded with this DAI - so pushing DAI up to the Debt Ceiling is potentially inefficient as if for example $100mm is injected in to the target protocol and that target protocol only has 10mm in under-collateralized loans to fund MD would only be receiving yield on $10 mill - which suggests that the parameter we should be targeting is not the debt ceiling - but instead the $ required to fund new approved loans in the target protocol - which is a number that in the case of Maple cannot be determined onchain 
- Would be great to see proposed code for how you plan to accommodate automation of any portfolio level constraints at the target protocol level - such as minimum debt service coverage or number of pool defaults that RWF would likely require in such a deal ?
- How do you reflect RWF requirements in the automated approach that may not be available at the target protocol on-chain - smart contracts whose semantic logic doesnt reflect the legal requirements of the deal that may be required by RWF ie... covenants or other required defaults ?
- how do you handle restrictions on withdrawals either due to explicit lockups and/or liquidity constraints that make it difficult to programmatically implement a withdrawal and potentially create a liquidity crunch within the target protocol?
- How do you avoid our D3M distorting market demand/supply dynamics - specifically providing liquidity where the reason for the D3M being used to provide liquidity to the target protocol - solely because there is not general market appetite to invest in the target protocol because the yields provided are mis-priced and/or the risk of the underlying asset is too high 
- how do we incorporate oracles if required to implement the automated trading approach ie. if public getters on the target protocols smart contracts dont provide the information ie. level of junior cover BPT price as an example?

hexonaut:
> how are fees accrued here? is it just that they send more dai than borrowed based on the agreement without an on-chain component?

Yes. This is the same as MIP21. Fees "accrue" on MIP21 continuously, but those are really a facade, and there is no enforcement that they will actually come in in the future beyond the agreement (which is the same here).

hexonaut:
> most of the MIP21s - I have reviewed so far have off-chain valuations for the underlying collateral - with the ERC20 token acting as a dummy placeholder, and not representing on-chain the true value of the underlying collateral - wondering how you would incorporate this in the plan?

From my understanding the oracle price is just set to the debt ceiling + accrued interest over some period of time. It is not an actual valuation of the collateral backing the debt. @godsflaw correct me if I'm wrong on that.

The Plan is no different in that we assume the collateral we can liquidate is worth at least the same as the debt position + interest.

> If D3M is about adjusting the amount of DAI invested to adjust the underlying demand/supply curve for DAI liquidity - not sure how this achieves that as it doesn't target a DAI borrowing rate but appears to target a level of underlying tokens that we want to hold?

MIP21 does not target any demand or supply curve either. This PR is purely to demonstrate that the D3M is a successor to MIP21.

> what is the incentive to ensure any DAI minted into the target protocol that this generates underlying yield for MDAO - specifically in the proposed Maple case - the lpTokens that MD would hold will only generate yield if there are actual loans funded with this DAI

Yes in Maple's case you only get interest if the DAI is used. This is getting outside the scope of this PR though which is purely the MIP21. We should discuss the Maple stuff separately.

And reading over the rest it seems like these are all about Maple. Let's discuss those outside this PR.

hexonaut:
Closing as this is not really relevant anymore.

PR #70, opened 2022-06-10
Introduce fix function and add tests

iamchrissmith:
adding additional test scenarios to v2

add additional test cases for debt paid back where:
- wind/unwind
- unwind to 0
- wind to DC
- reap

should add tests that:

1. cages d3m
2. partial unwind
3. dss gets caged
4. d3m gets liquidity
5. unwind continues

cull/uncull

1. cages d3m
2. partial unwind
3. cull
4. dss gets caged
5. uncull
6. d3m gets liquidity
7. unwind continues


talbaneth:
@iamchrissmith this is based of `v2` right? can you fix that on this github pr page?

iamchrissmith:
> @iamchrissmith this is based of `v2` right? can you fix that on this github pr page?

It is, but GH won't let me change the base until it differs from v2, so i can't until i push up what i have.  This is still draft and not ready to review yet. also these are just tests so not critical to have frozen for the audit we said

hexonaut:
When the assertRevert is added on the other PR these tests should be converted from the `testFail` anti-pattern to asserting the actual revert messages.

iamchrissmith:
> When the assertRevert is added on the other PR these tests should be converted from the `testFail` anti-pattern to asserting the actual revert messages.

maybe i'm missing the point of your comment, but yes all the changes Gonza made will be merged in here

hexonaut:
> > When the assertRevert is added on the other PR these tests should be converted from the `testFail` anti-pattern to asserting the actual revert messages.
> 
> maybe i'm missing the point of your comment, but yes all the changes Gonza made will be merged in here

Nevermind. I thought you added some more.

src/D3MHub.sol
```diff
@@ -380,6 +382,23 @@ contract D3MHub {
         return ilks[ilk].tic;
     }
 
+    /**
+        @notice Rebalance position if someone permissionlesly paid part of the debt
+        @dev This function will regenerate the debt of the position in case art < ink
+        and send the fresh DAI generated to the surplus buffer.
+        @param ilk bytes32 of the D3M ilk name
+    */
+    function fix(bytes32 ilk) external lock {
```
naszam, line 391: would this require a `vat.live` check (as `suck` is being used)?

iamchrissmith: i kind of think we would still want this right after cage if we are unbalanced.  Perhaps we add a `require(end.debt() == 0)` check to we don't suck after the outstanding dai is accounted for?

naszam: yeah, after a second look, I might see why it wasn't added but maybe as you pointed out a safeguard could be considered here.

naszam: A `vat.live` check was introduced in 8eb2048, going to resolve.

src/D3MHub.sol
```diff
@@ -255,7 +255,9 @@ contract D3MHub {
         if (mode == Mode.NORMAL) {
             // Normal mode or module just caged (no culled)
             // debt is obtained from CDP art
-            (,daiDebt) = vat.urns(ilk, address(_pool));
+            uint256 ink;
+            (ink, daiDebt) = vat.urns(ilk, address(_pool));
+            require(ink == daiDebt, "D3MHub/position-needs-to-be-fixed");
```
talbaneth, line 260: I think this can be removed.
If it was left as an internal sanity check then it's probably redundant as we call `_fix()` before the two places where `Mode.NORMAL` is passed to unwind.

src/D3MHub.sol
```diff
@@ -331,6 +333,16 @@ contract D3MHub {
         emit Unwind(ilk, amount, fees);
     }
 
+    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
+        (uint256 ink, uint256 art) = vat.urns(ilk, _pool);
+        if (art < ink) {
+            address _vow = vow;
+            diff = _min(ink - art, MAXINT256);
```
talbaneth, line 340: If the diff is somehow greater than `MAXINT256` we should probably revert, as even if we fix `MAXINT256` we keep our invariant of `ink==art` broken.
I think such a big diff can only be in some attack scenario, in which case we probably want to block `exec` anyway and have governance try to handle the case.

src/tests/interfaces/interfaces.sol
```diff
@@ -43,6 +43,10 @@ interface EndLike {
     function cage(bytes32) external;
     function skim(bytes32, address) external;
     function thaw() external;
+    function gap(bytes32) external returns (uint256);
+    function cash(bytes32, uint256) external;
+    function pack(uint256) external;
+    function flow(bytes32) external;
```
talbaneth, line 49:
```suggestion
```
unused.

src/tests/integration/D3MAave.t.sol
```diff
@@ -1061,6 +1061,86 @@ contract D3MAaveTest is DSSTest {
         vat.frob(ilk, address(address(d3mAavePool)), address(this), address(this), 0, -100); // Some small amount of dai repaid
 
         // We should be able to close out the vault completely even though ink and art do not match
-        _setRelBorrowTarget(0);
+        // _setRelBorrowTarget(0);
+        d3mAavePlan.file("bar", 0);
+
+        d3mHub.exec(ilk);
```
talbaneth, line 1067: Consider adding pool `ink` and `art` checks before and after this call.

src/tests/integration/D3MAave.t.sol
```diff
@@ -1061,6 +1061,86 @@ contract D3MAaveTest is DSSTest {
...
+    }
+
+    function test_wind_partial_unwind_wind_debt_paid_back() public {
```
talbaneth, line 1070: We should clean up some leftover comments in this test.

src/tests/D3MHub.t.sol
```diff
@@ -395,6 +399,264 @@ contract D3MHubTest is DSSTest {
...
+        assertEq(art, 75 * WAD);
+        assertEq(vat.gem(ilk, address(d3mTestPool)), 0);
+        assertEq(vat.vice(), viceBefore);
+        assertEq(vat.sin(vow), sinBefore);
+        assertEq(vat.dai(vow), vowDaiBefore + 10 * RAD);
+        assertEq(dai.balanceOf(address(testGem)), 75 * WAD);
+        assertEq(testGem.balanceOf(address(d3mTestPool)), 75 * WAD);
+    }
+
+    function test_wind_fully_unwind_wind_debt_paid_back() public {
```
talbaneth, line 500:
```suggestion
    function test_fully_unwind_after_debt_paid_back() public {
```
to be consistent with the previous test's name.

talbaneth (commented):
Finished reviewing the changes and tests. See suggestions.

src/D3MHub.sol
```diff
@@ -331,6 +333,16 @@ contract D3MHub {
         emit Unwind(ilk, amount, fees);
     }
 
+    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
```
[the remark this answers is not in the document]
gbalabasquer: It's the same net effect in a bit more optimized way. If you apply `fix` first it will increase the debt and the surplus. Which will be balanced out when grabbing the position.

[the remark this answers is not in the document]
gbalabasquer: I just deleted my prev comment as I think you might be right, need to think a bit in the scenarios and will get back.

src/D3MHub.sol
gbalabasquer: Yeah totally, I had forgotten to remove it, it's done now.

src/D3MHub.sol
gbalabasquer: Actually I think we should remove the `_min` at all. In `_wind` there is a protection to not leave the position > MAXINT256, so if there isn't a contract bug is impossible to get to that state. We should also be able to remove it from `_unwind`.

src/tests/integration/D3MAave.t.sol
gbalabasquer: If you meant the `vat.sin` ones were commented due stack too deep. But seems Chris did actually replicate some variable cleaning in the integration tests and they can now be enabled (which is done now + adding `vice` checks)

src/tests/integration/D3MAave.t.sol
gbalabasquer: I've modified substantially this test, let me know if looks good now.

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: I actually think we don't have to use `fix` in `cull` at all but also removing the actual rebalancing from it.
We originally stated that if `cull` happens we should only take the real actual debt as the `gem` to keep that reference. However if someone donated DAI paying back this CDP in reality is donating DAI to the surplus buffer, so maybe the concept of leaving the reference to the actual art is actually wrong, it should always been the actual `ink`.

If we remove the actual rebalancing, calling `fix` or not is the exact same net result, as whatever we send to the surplus buffer in `fix` will be balanced out when the `cull` `grab` is called.

src/D3MHub.sol
```diff
@@ -560,15 +576,6 @@ contract D3MHub {
         require(art <= MAXINT256, "D3MHub/overflow");
         vat.grab(ilk, address(_pool), address(_pool), vow, -int256(ink), -int256(art));
 
-        if (ink > art) {
```
naszam, line 563: just a reminder to update the comments in the `cull` NatSpec and consider documenting `fix` behaviour as well in `exec`

src/D3MHub.sol
```diff
@@ -331,6 +331,16 @@ contract D3MHub {
         emit Unwind(ilk, amount, fees);
     }
 
+    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
```
naszam, line 334: A note on the naming, if we're going with the change I think some daiwanese could be explored, consider something like `mend`, `give`, `neat`, `tidy`, `keel`.
I personally like the last one, `keel` as it seems like a [special one](https://en.wikipedia.org/wiki/Keel), it comes from the Latin `Carina`, is the origin of the term `careen`, to clean a keel and regarded by some scholars as the first word in the English language recorded in writing.

hexonaut (approved):
Agree with removing the `min` in `_fix()`, but rest looks good. Good test coverage.

src/D3MHub.sol
talbaneth: yeah, it can be removed.

resource_errors.json
```diff
@@ -0,0 +1,3 @@
+{
+    "topics": []
+}
```
naszam, line 3: artifact, consider adding it to `.gitignore`

src/tests/integration/D3MAave.t.sol
talbaneth: ok.

src/D3MHub.sol
talbaneth: Always up for daiwanese but don't forget it's an internal function.

src/D3MHub.sol
gbalabasquer: Comment fixed!

resource_errors.json
gbalabasquer: Done!

src/D3MHub.sol
gbalabasquer: I had made a long comment about it and not sure what happened, if was removed somehow, or not even sent.
Anyway what I was saying is that I'm actually not sure what to do it here. In theory as we are blocking in `wind` to be `> MAXINT256`, then we should be able to remove any other check. However that is relying 100% than an external source won't alter it, the invariant can be broken as the storage doesn't happen here but in the `vat`.
Thinking on that is nice to have checks that ensure no overflow can happen at all, just as a best practice. That can be done with `require`s as well but we had decided to remove it from `unwind` to avoid any case of stopping it.
Putting it here, kind of breaks that reasoning, so I think we have these different scenarios:
1) Remove all the checks at all and rely an external source won't make this invariant to break
2) Add just requires for best practice as they won't trigger anyway if there isn't an external intervention.
3) Leave the min and in case it would trigger (which shouldn't happen without external intervention) the imbalance will be fixed partially.

I think I prefer option 2.

talbaneth: agreeing on 2, that was my original preference.

src/D3MHub.sol
naszam: yeah, I shared for discussion in case we'd like to explore alternative naming.
If we'd like to stick with a 3 letters for an internal function, maybe `pop` could be an alternative.

src/D3MHub.sol
gbalabasquer: Done: https://github.com/makerdao/dss-direct-deposit/pull/70/commits/677306e6971858795f7a21cfb3fb41505a95f732

talbaneth: looks good, resolving this.

src/D3MHub.sol
```diff
@@ -490,6 +503,8 @@ contract D3MHub {
         _pool.preDebtChange("reap");
         uint256 assetBalance = _pool.assetBalance();
         (, uint256 daiDebt) = vat.urns(ilk, address(_pool));
+        daiDebt += _fix(ilk, address(_pool));
```
gbalabasquer, line 506: There is another way to do this which is first executing `_fix`, then calling vat.urns and get the `daiDebt` corrected.
We could also change `_fix` to fixed daiDebt and in this case we would save two warmed SLOADs.
For `exec` though is better to get the diff as we are adding up to the `Art`. Conceptually it shouldn't be correct to replace the `Art` by the `art` of one CDP regardless it is the unique one of this type.

src/D3MHub.sol
```diff
@@ -331,6 +329,17 @@ contract D3MHub {
         emit Unwind(ilk, amount, fees);
     }
 
+    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
```
gbalabasquer, line 332: Should we explore the idea to have multiple ways to enter this function to avoid loading the vat.urns multiple times? I don't think it is worth it due gas is not sensitive here but just want to make sure we agree on this.

hexonaut: I noticed this gas optimization, but didn't comment because:

1. These functions are not gas sensitive. Infrequently called for large sums of money. The gas is trivial.
2. The SLOAD is already hot, so it's not that bad.

hexonaut (approved):
New changes look good.

src/D3MHub.sol
hexonaut: As I said in the other comment, I'm not too concern about gas optimizations here.

talbaneth (approved):
LGTM

src/tests/integration/D3MAave.t.sol
```diff
@@ -1050,17 +1050,141 @@ contract D3MAaveTest is DSSTest {
...
 
-        // Permissionlessly repay the urn
-        _giveTokens(dai, 100);
-        dai.approve(address(daiJoin), 100);
-        daiJoin.join(address(this), 100);
-        vat.frob(ilk, address(address(d3mAavePool)), address(this), address(this), 0, -100); // Some small amount of dai repaid
+        d3mHub.reap(ilk); // Clear out fees at the start
+
+        (uint256 pink, uint256 part) = vat.urns(ilk, address(d3mAavePool));
+        uint256 gemBefore = vat.gem(ilk, address(end));
```
naszam, line 1061:
```suggestion
        uint256 gemBefore = vat.gem(ilk, address(d3mAavePool));
```

src/tests/integration/D3MAave.t.sol
```diff
@@ -1050,17 +1050,141 @@ contract D3MAaveTest is DSSTest {
...
+        assertEq(dai.balanceOf(address(adai)), adaiDaiBalanceInitial);
+        assertEq(adai.balanceOf(address(d3mAavePool)), 0);
+    }
+
+    function test_wind_partial_unwind_wind_debt_paid_back() public {
+        uint256 initialRate = _setRelBorrowTarget(5000);
+        d3mHub.reap(ilk); // Clear out fees at the start
+
+        (uint256 pink, uint256 part) = vat.urns(ilk, address(d3mAavePool));
+        uint256 gemBefore = vat.gem(ilk, address(end));
```
naszam, line 1114:
```suggestion
        uint256 gemBefore = vat.gem(ilk, address(d3mAavePool));
```

PR #71, opened 2022-06-13
Overriding fail tests

talbaneth:
With the current deposit/withdraw/transfer/quit fail tests in base test contract, some of them do not actually set the initial conditions for the test to pass (e.g testFail_deposit_not_hub fails even if hub is relied in the pool).
Fixing that needs aave specific code (`D3MTestGem(address(adai)).rely(address(aavePool));`).
Same situation exists with Compound for some of these fail tests.

Therefore I made these tests virtual in base, and moved the implementation back to D3MAaavePool.t.sol.
This way we don't forget to implement these tests for each platform.


gbalabasquer:
I think this one can be closed now.

PR #72, opened 2022-06-13
Assert revert

gbalabasquer:


hexonaut (commented):
I'm okay with pulling in just part of `dss-test`, but it should be almost entirely backwards compatible. It may be easier to just import `dss-test` ?

talbaneth:
I'm fine with either way (`dss-test` / standalone file) as long as we keep dapptools compatibility for now.

gbalabasquer:
PR is ready for review. I'm fine adding `assertRevert` to `dss-test` and replacing it. Just wanted to avoid forcing dapptools deprecation at this point and felt easier to include the standalone fine in the short term. But if `dss-test` can incorporate it very soon, then we can do the replacement.

src/tests/DssTest.sol
```diff
@@ -0,0 +1,53 @@
...
+                mstore(0x40, add(message, and(add(add(size, 0x20), 0x1f), not(0x1f))))
+                returndatacopy(add(message, 0x20), 0x44, size)
+            }
+            if (keccak256(abi.encodePacked(message)) != keccak256(abi.encodePacked(expectedMessage))) {
+                emit log("Error: revert message not satisfied");
+                fail();
+            }
+        }
+    }
+}
```
naszam, line 53: missing eof newline

hexonaut:
> PR is ready for review. I'm fine adding `assertRevert` to `dss-test` and replacing it. Just wanted to avoid forcing dapptools deprecation at this point and felt easier to include the standalone fine in the short term. But if `dss-test` can incorporate it very soon, then we can do the replacement.

`dss-test` already supports `assertRevert`: https://github.com/makerdao/dss-test/blob/master/src/DSSTest.sol#L89

hexonaut (approved):
I'm okay if you want to just leave this `DssTest` instead of importing the library. Good to merge.

src/tests/DssTest.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-FileCopyrightText: © 2021-2022 Dai Foundation <www.daifoundation.org>
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2021-2022 Dai Foundation
```
naszam, line 3:
```suggestion
// Copyright (C) 2021 Dai Foundation
```

src/D3MHub.t.sol
```diff
@@ -901,21 +898,21 @@ contract D3MHubTest is DSTest {
...
 
-    function testFail_no_uncull_mcd_live() public {
+    function test_no_uncull_mcd_live() public {
         _windSystem();
         d3mHub.cage(ilk);
 
         d3mHub.cull(ilk);
 
-        d3mHub.uncull(ilk);
+        assertRevert(address(d3mHub), abi.encodeWithSignature("uncull(bytes32)", ilk), "D3MHub/no-uncull-normal-operation");
```
talbaneth, line 911: In `testFail_hub_upgrade_kills_old_hub` you can change to `test_hub_upgrade_kills_old_hub` and do:

```
        // Ensure old hub revert
        (bool succeeded,) = address(d3mHub).call(abi.encodeWithSignature("exec(bytes32)", ilk));
        assert(!succeeded);
```
This will remove the last failTest.

gbalabasquer: `assertRevert` actually implements this way. Or am I missing anything?

talbaneth: Your are currently not using `assertRevert` for `testFail_hub_upgrade_kills_old_hub`.
I figured it was because there is no revert message there.
If that's the case then you can just do that instead.
Did not see an `assertRevert` version that does not check the error message.

gbalabasquer: I tried using assertRevert but when running the test I got a failure due OutOfGas and not the missmatching error message. So like something is breaking getting outside the `call` environment. This is happening at least with Foundry.

talbaneth: It's probably becuase it's calling a view functions on address(0) (we set the pool to that in that test).

gbalabasquer: Ah yes that might explain it, but want to do some deeper check later. For now I just left that one remaining test as it was.

src/tests/D3MOracle.t.sol
```diff
@@ -139,11 +137,8 @@ contract D3MOracleTest is DSTest {
         assertEq(oracle.read(), WAD);
         vat.cage();
         assertEq(oracle.read(), WAD);
-    }
-
-    function testFail_read() public {
         hub.cull();
         vat.cage();
-        oracle.read();
+        assertRevert(address(oracle), abi.encodeWithSignature("read()"), "D3MOracle/ilk-culled-in-shutdown");
```
talbaneth, line 142: Nice!

iamchrissmith (approved):
thank you for making all these updates!

talbaneth (approved):
LGTM.
The open comment about avoiding the last `testFail` is not critical at all.

PR #74, opened 2022-06-15
TrueFi V1 Support

MiksuJak:


hexonaut:
Thanks guys. Will look at this when we get a chance. Still finishing up with v2 rewrite and Compound D3M.

hexonaut:
Can this be updated to point at `master` and re-opened? I can't seem to do that myself.

iamchrissmith:
> Can this be updated to point at `master` and re-opened? I can't seem to do that myself.

done

PR #83, opened 2022-08-03
Remove reap, simplify _unwind and rebalance

gbalabasquer:


src/D3MHub.sol
```diff
@@ -221,115 +221,63 @@ contract D3MHub {
...
-        emit Unwind(ilk, amount, fees);
+        emit Unwind(ilk, amount);
     }
 
-    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
+    function _fix(bytes32 ilk, address _pool, uint256 assets) internal returns (uint256 diff) {
         (uint256 ink, uint256 art) = vat.urns(ilk, _pool);
+        if (assets > ink) {
+            uint256 _diff = assets - ink;
+            require(_diff <= MAXINT256, "D3MHub/overflow");
```
iamchrissmith, line 276:
```suggestion
            _diff = min(_diff, MAXINT256);
```

This way if we ever end up in this situation we can fix it with two `_fix`es (i think?)

iamchrissmith (changes requested):

src/D3MHub.sol
```diff
@@ -221,115 +221,63 @@ contract D3MHub {
...
-            daiDebt = vat.gem(ilk, address(_end));
-        } else revert("D3MHub/unknown-mode");
-
-        uint256 availableAssets = _pool.maxWithdraw();
-
-        // Unwind amount is limited by how much:
-        // - max reduction desired
-        // - assets available
-        // - dai debt tracked in vat (CDP or free)
+    function _unwind(bytes32 ilk, ID3MPool _pool, uint256 reduction, Mode mode, uint256 assetBalance) internal {
```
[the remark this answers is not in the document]
gbalabasquer: I took the liberty to remove it to keep consistency with `_wind` where we do not do it for `amount`.
Maybe we can clarify it differently?

src/D3MHub.sol
gbalabasquer: Well fixing in two steps might bring unexpected consequences. As for example we won't be able to ensure anymore that `assetsBalance == art` in normal operation.
Trying to remember what was the conclusion we had arrived here as `_fix` already had the last overflow check.

src/D3MHub.sol
```diff
@@ -416,7 +368,9 @@ contract D3MHub {
                 currentAssets
             );
         } else if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active()) {
-            _fix(ilk, address(_pool));
+            if (ilks[ilk].culled == 0) {
```
talbaneth, line 371: `lks[ilk].culled` can be read to a local variable here instead of being read twice from storage.

src/D3MHub.sol
```diff
@@ -221,115 +221,64 @@ contract D3MHub {
...
+        emit Unwind(ilk, amount);
     }
 
-    function _fix(bytes32 ilk, address _pool) internal returns (uint256 diff) {
+    function _fix(bytes32 ilk, address _pool, uint256 assets) internal returns (uint256 diff) {
         (uint256 ink, uint256 art) = vat.urns(ilk, _pool);
+        if (assets > ink) {
+            uint256 _diff = assets - ink;
+            require(_diff <= MAXINT256, "D3MHub/overflow");
+            vat.slip(ilk, address(_pool), int256(_diff));
```
talbaneth, line 277: `_pool` doesn't need casting to address.

src/D3MHub.sol
hexonaut: It seems to me A) this will never happen in practice and B) we probably want to just revert for a simpler mental model

hexonaut (approved):
Looks good to me. Really like these refactors it's much cleaner.

hexonaut (changes requested):
Tests need to be fixed.

src/D3MHub.sol
```diff
@@ -221,121 +219,123 @@ contract D3MHub {
...
+            if (debt > Line) {
+                unchecked {
+                    toUnwind = _max(toUnwind, _divup(debt - Line, RAY)); // checks if we need to unwind due global debt ceiling
+                }
+            }
 
-        emit Unwind(ilk, amount, fees);
-    }
+            // Determine if it needs to unwind due plan
+            uint256 targetAssets = ilks[ilk].plan.getTargetAssets(currentAssets);
```
talbaneth, line 306: `getTargetAssets` appears twice in the code (and is called twice if we end up in the `wind` path).
Can we store it in a local var that's declared near `toUnwind`? (tried it locally and did not encounter stack too deep problems).

src/D3MHub.sol
```diff
@@ -221,121 +219,123 @@ contract D3MHub {
...
-            return;
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(_pool));
+        uint256 maxWithdraw = _pool.maxWithdraw();
+        if (currentAssets > ink) { // If fees were generated
+            uint256 fixInk = _min(
+                currentAssets - ink, // fees generated
+                ink <= lineWad // if previously CDP was under debt ceiling
+                    ? lineWad + maxWithdraw - ink // we can just go above debt ceiling + maxWithdraw
+                    : maxWithdraw // we can just go above current state + maxWithdraw
+            );
```
talbaneth, line 268: Subjective, but you can consider writing it like this (maybe easier to understand):

```
uint256 underLine = (ink <= lineWad) ? (lineWad - ink) : 0;
uint256 fixInk = _min(currentAssets - ink, underLine + maxWithdraw)
```

src/D3MHub.sol
```diff
@@ -395,128 +395,47 @@ contract D3MHub {
...
-                currentAssets
+                address(_end)
             );
-        } else if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active()) {
-            _fix(ilk, address(_pool));
-
+        } else if (ilks[ilk].culled == 1) {
             // pool caged
-            _unwind(
+            _fullUnwind(
```
talbaneth, line 422: The `pool caged` comment above should change to `pool culled` or dropped.

src/D3MHub.sol
```diff
@@ -395,128 +395,47 @@ contract D3MHub {
...
+        // IMPORTANT: this function assumes Vat rate of D3M ilks will always be == 1 * RAY (no fees).
+        // That's why this module converts normalized debt (art) to Vat DAI generated with a simple RAY multiplication or division
+
         (uint256 Art, uint256 rate, uint256 spot, uint256 line,) = vat.ilks(ilk);
         require(rate == RAY, "D3MHub/rate-not-one");
         require(spot == RAY, "D3MHub/spot-not-one");
 
         ID3MPool _pool = ilks[ilk].pool;
 
         _pool.preDebtChange("exec");
```
talbaneth, line 407: Not a major thing, but since we removed `reap` we should consider removing the "exec" identifier in `pool.preDebtChange()` and `pool.postDebtChange()`, as these are callend only in `exec()` now.

src/D3MHub.sol
```diff
@@ -221,121 +219,123 @@ contract D3MHub {
...
+            require(fixInk <= MAXINT256, "D3MHub/overflow");
+            vat.slip(ilk, address(_pool), int256(fixInk)); // Generate extra collateral
+            vat.frob(ilk, address(_pool), address(_pool), address(this), int256(fixInk), 0); // Lock it
+            ink += fixInk;
+            emit Fees(ilk, fixInk);
+        }
+        if (art < ink) { // If there was permissionless DAI paid or fees added as collateral
+            address _vow = vow;
+            uint256 fixArt = ink - art; // Amount of fees + permissionless DAI paid we will now transform to debt
+            delete art;
```
talbaneth, line 278: Is this needed?

hexonaut (commented):
Overall I'm liking this refactor even more. Code looks functionally the same with the addition of the `line + maxWithdraw` interest earned limit. Just two comments on non-critical changes.

src/D3MHub.sol
hexonaut: Maybe a comment to show the expanded form so we don't increase gas? I agree it's a little difficult to understand what's going on here.

src/D3MHub.sol
```diff
@@ -228,63 +226,116 @@ contract D3MHub {
...
+            toUnwind = _min(
+                toUnwind,
+                maxWithdraw
+            );
+            require(toUnwind <= MAXINT256, "D3MHub/overflow");
+            _unwind(ilk, _pool, toUnwind);
+        } else {
+            uint256 toWind;
+            // All the subtractions are safe as otherwise toUnwind is > 0
+            uint256 targetAssets = ilks[ilk].plan.getTargetAssets(currentAssets);
```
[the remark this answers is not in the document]
hexonaut: Just realized Tal commented on this already. If there is no stack too deep then fully in favour of caching.

src/D3MHub.sol
gbalabasquer: I've re-structured this further so this should be solved now.

src/D3MHub.sol
gbalabasquer: Same here, this should be solved now.

src/D3MHub.sol
gbalabasquer: Fixed!

src/D3MHub.sol
gbalabasquer: Comment removed, it is pretty much self explanatory

src/D3MHub.sol
gbalabasquer: Good catch, it's done now!

src/D3MHub.sol
gbalabasquer: Not really, I put it just to see it helped with stack too deep (that I used to get), but mainly to show that variable is discarded. As the `ink` and `Art` are updated, it felt a bit weird leaving the `art` non updated, but also didn't want to update it if not used anymore.

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: Yes, this is fixed now.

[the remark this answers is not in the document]
gbalabasquer: There is only one case, when actually now we compute it and is not needed. The case where the `pool` is caged or inactive. But felt it didn't deserve adding conditions to make it more gas efficient for this particular case.

src/D3MHub.sol
```diff
@@ -228,63 +226,116 @@ contract D3MHub {
...
-            emit Fees(ilk, _diff);
+    function _normal(bytes32 ilk, ID3MPool _pool, uint256 Art, uint256 lineWad) internal {
+        uint256 currentAssets = _pool.assetBalance(); // Should return DAI owned by D3MPool
+
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(_pool));
+        uint256 maxWithdraw = _pool.maxWithdraw();
+        if (currentAssets > ink) { // If fees were generated
+            uint256 fixInk = _min(
+                currentAssets - ink, // fees generated
+                ink <= lineWad // if previously CDP was under debt ceiling
```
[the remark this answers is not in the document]
gbalabasquer: Fixed.

src/D3MHub.sol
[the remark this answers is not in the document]
talbaneth: I actually think you shouldn't call `getTargetAssets` or rely on its result if the pool is not active (as we don't rely on the implementation anymore).

src/D3MHub.sol
```diff
@@ -221,121 +219,122 @@ contract D3MHub {
...
+            vat.slip(ilk, address(_pool), int256(fixInk)); // Generate extra collateral
+            vat.frob(ilk, address(_pool), address(_pool), address(this), int256(fixInk), 0); // Lock it
+            ink += fixInk;
+            emit Fees(ilk, fixInk);
+        }
+        if (art < ink) { // If there was permissionless DAI paid or fees added as collateral
+            address _vow = vow;
+            uint256 fixArt = ink - art; // Amount of fees + permissionless DAI paid we will now transform to debt
+            delete art;
+            Art += fixArt;
```
talbaneth, line 279: Not sure if we usually change the values of input variables like this.
Do we have enough place on the stack for another local var?

(this and the `delete art` line make me think if we might want the _fix part as a separate function after all).

gbalabasquer: Ah right, now the `Art` is a param. Yeah in theory we shouldn't be touching this variable. Let me see if I can do something better.

src/D3MHub.sol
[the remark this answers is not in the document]
talbaneth: (for example it can revert and not allow you to unwind).

src/D3MHub.sol
hexonaut: Was not even aware of this. You can free up stack slots with `delete`...? 🤯

talbaneth: I don't think you can, iirc it just assigns 0 to it.

gbalabasquer: Yeah yeah, not sure if actually does something about the stack too deep. Probably not as Tal says.

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: I think now is fine. I've also set back the `toUnwind = type(uint256).max` to make sure we enter the unwind path even if maxWithdraw is 0.

hexonaut (approved):
Looks good now pending Tal's comments.

src/D3MHub.sol
```diff
@@ -221,121 +219,121 @@ contract D3MHub {
...
-            }
+            _pool.withdraw(amount);
+            daiJoin.join(address(this), amount);
 
-            if (amount + fees > availableAssets) {
-                unchecked {
-                    fees = availableAssets - amount;
-                }
-            }
+            vat.slip(ilk, urn, -int256(toSlip));
```
talbaneth, line 251: Consider moving the `vat.slip` to after the require 3 lines above. 
Will separate nicely between the `toSlip` and `amount` usage.

src/D3MHub.sol
```diff
@@ -221,121 +219,121 @@ contract D3MHub {
...
-        // IMPORTANT: this function assumes Vat rate of D3M ilks will always be == 1 * RAY (no fees).
-        // That's why this module converts normalized debt (art) to Vat DAI generated with a simple RAY multiplication or division
-        // This module will have an unintended behaviour if rate is changed to some other value.
-        if (amount == 0) {
-            emit Wind(ilk, 0);
-            return;
+        if (amount > 0) {
+            vat.slip(ilk, address(_pool), int256(amount));
+            vat.frob(ilk, address(_pool), address(_pool), address(this), int256(amount), int256(amount));
+            // normalized debt == erc20 DAI (Vat rate for D3M ilks fixed to 1 RAY)
```
talbaneth, line 225: Can we remove this comment? it visually breaks the nice symmetry with `_unwind`.

src/D3MHub.sol
```diff
@@ -221,121 +219,121 @@ contract D3MHub {
...
-            vat.move(address(this), vow, total * RAY);
+        if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active()) { // If D3M is caged (but not culled) or plan is not active
+            toUnwind = type(uint256).max; // We make sure to enter the unwind path
         } else {
-            // This can be done with the assumption that the price of 1 collateral unit equals 1 DAI.
-            // That way we know that the prev End.skim call kept its gap[ilk] emptied as the CDP was always collateralized.
-            // Otherwise we couldn't just simply take away the collateral from the End module as the next line will be doing.
-            vat.slip(ilk, address(_end), -int256(amount));
-            vat.move(address(this), vow, total * RAY);
+            if (art > lineWad) {
```
talbaneth, line 294: So about the lack of symmetry, I guess there are 3 things that bothered me, and probably not much we want to change about them:

1. The winding check is more compact - we just do _min(_min(_min.. since we already know all the substractions are safe.
2. There is no check of `if (toWind > 0)` like we have for `toUnWind`, since we go to the wind case even if the amount is 0 (so we can emit an event).
3. `getTargetAssets` is called as part of the unwind checks although it is relevant to both unwind and wind.

(1) and (2) are fine, regarding (3), you might say it's completely redundant but we can do:

```
		if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active()) { // If D3M is caged (but not culled) or plan is not active
			toUnwind = type(uint256).max; // We make sure to enter the unwind path
		} else {
			targetAssets = ilks[ilk].plan.getTargetAssets(currentAssets);
		}

		if (toUnwind == 0) {
			if (art > lineWad) {
....
```

Not insisting on it in any case.

src/D3MHub.sol
```diff
@@ -221,121 +219,121 @@ contract D3MHub {
...
+            vat.slip(ilk, urn, -int256(toSlip));
+            vat.move(address(this), vow, amount * RAY);
         }
+        emit Unwind(ilk, amount);
+    }
 
-        if (amount == 0 && fees == 0) {
-            emit Unwind(ilk, 0, 0);
-            return;
+    function _normal(bytes32 ilk, ID3MPool _pool, uint256 Art, uint256 lineWad) internal {
```
naszam, line 257: Daiwanese attempt: `glad`
```suggestion
    function _glad(bytes32 ilk, ID3MPool _pool, uint256 Art, uint256 lineWad) internal {
```

src/D3MHub.sol
```diff
@@ -221,121 +219,121 @@ contract D3MHub {
...
-        // - dai debt tracked in vat (CDP or free)
-        uint256 amount = _min(
-                            _min(
-                                supplyReduction,
-                                availableAssets
-                            ),
-                            daiDebt
-                        );
-        require(amount <= MAXINT256, "D3MHub/overflow");
+    function _fullUnwind(bytes32 ilk, ID3MPool _pool, address urn) internal {
```
naszam, line 242: Daiwanese attempt: `unfurl`
```suggestion
    function _unfurl(bytes32 ilk, ID3MPool _pool, address urn) internal {
```

src/D3MHub.sol
gbalabasquer: Done!

src/D3MHub.sol
gbalabasquer: Yes I saw this one during the walkthrough and was thinking the same. Removed now!

src/D3MHub.sol
gbalabasquer: I prefer `_normal` to be honest. At the end they are internal functions and not even all the external are respecting daiwanese. We have `uncull` for example.

src/D3MHub.sol
gbalabasquer: This is totally forced and not even daiwanese. I proceeded to just overload the method `unwind`.

src/D3MHub.sol
naszam: what about `norm` to shorten a bit the function name?

src/D3MHub.sol
naszam: yeah, the attempt was more to reduce name size with a synonym of `unwind`, overloading the method sounds better.

src/tests/integration/D3MAave.t.sol
```diff
@@ -529,39 +535,58 @@ contract D3MAaveTest is DSSTest {
...
+        assertEqApprox(vat.dai(vow), vowDai + feesAccrued * RAY, RAY);
 
         // Someone repays
         aavePool.repay(address(dai), amountToBorrow, 2, address(this));
         d3mHub.exec(ilk);
 
+        // Now the CDP completely unwinds and surplus buffer doesn't change
         (ink, art) = vat.urns(ilk, address(d3mAavePool));
         assertEq(ink, 0);
         assertEq(art, 0);
```
talbaneth, line 554:
```suggestion
        assertEqApprox(ink, 0, 1);
        assertEqApprox(art, 0, 1);
```

Fixes to failure of this test on block 15308303.

src/D3MHub.sol
gbalabasquer: Made the change discussed in the call. If feels good this way we can get this comment solved.

src/D3MHub.sol
gbalabasquer: Pushed a minor change, let me know if feels better now.

talbaneth: yes

src/tests/D3MHub.t.sol
```diff
@@ -1699,6 +1640,263 @@ contract D3MHubTest is DSSTest {
...
+
+        d3mHub.exec(ilk);
+
+        (ink, art) = vat.urns(ilk, address(d3mTestPool));
+        assertEq(ink, 50 * WAD);
+        assertEq(art, 50 * WAD);
+        assertEq(vat.dai(vow), prevDai + 3 * RAD);
+        assertEq(testGem.balanceOf(address(d3mTestPool)), 57 * WAD);
+    }
+
```
talbaneth, line 1787: Should we add the case for ink==line and maxWithdraw > fees?
(ink: 50, fees: 10, line: 50, to withdraw > 10).

src/tests/D3MHub.t.sol
```diff
@@ -1699,6 +1640,263 @@ contract D3MHubTest is DSSTest {
...
+        uint256 prevDai = vat.dai(vow);
+
+        d3mHub.exec(ilk);
+
+        (ink, art) = vat.urns(ilk, address(d3mTestPool));
+        assertEq(ink, 50 * WAD);
+        assertEq(art, 50 * WAD);
+        assertEq(vat.dai(vow), prevDai + 10 * RAD);
+        assertEq(testGem.balanceOf(address(d3mTestPool)), 50 * WAD);
+    }
```
talbaneth, line 1666: Do you think it's worth it to add a second exec call to these tests and make sure nothing has changed?

src/D3MHub.sol
```diff
@@ -221,121 +219,125 @@ contract D3MHub {
...
+                unchecked {
+                    toUnwind = _max(toUnwind, _divup(debt - Line, RAY));
+                }
+            }
+            // Determine if it needs to unwind due plan targetAssets
+            if (targetAssets < currentAssets) {
+                unchecked {
+                    toUnwind = _max(toUnwind, currentAssets - targetAssets);
+                }
+            }
```
talbaneth, line 314: Consider putting all of this under one `unchecked` block.
Will make the code more compact and visually more similar to the winding logic below.

hexonaut: As much as it is a little verbose, I don't like having too large of unchecked blocks as your default mindset is that everything is checked. This exception seems small enough that it's probably okay though.

src/D3MHub.sol
```diff
@@ -239,32 +238,31 @@ contract D3MHub {
...
             _pool.withdraw(amount);
             daiJoin.join(address(this), amount);
-
-            vat.slip(ilk, urn, -int256(toSlip));
             vat.move(address(this), vow, amount * RAY);
         }
         emit Unwind(ilk, amount);
     }
 
     function _normal(bytes32 ilk, ID3MPool _pool, uint256 Art, uint256 lineWad) internal {
```
hexonaut, line 255: Why is `Art` being passed in if it's not being used?

src/D3MHub.sol
```diff
@@ -275,38 +273,43 @@ contract D3MHub {
...
+            targetAssets = ilks[ilk].plan.getTargetAssets(currentAssets);
+
+            // Determine if it needs to unwind due ilk debt ceiling
+            if (art > lineWad) {
                 unchecked {
-                    toUnwind = Art - lineWad; // checks if we need to unwind due ilk debt ceiling
+                    toUnwind = art - lineWad;
                 }
             }
+            // Determine if it needs to unwind due global debt ceiling
```
hexonaut, line 303: Minor English - should be "due to" in all these comments

src/tests/integration/D3MCompound.t.sol
```diff
@@ -96,6 +96,9 @@ contract D3MCompoundTest is DSSTest {
     uint256 constant INTEREST_RATE_TOLERANCE = WAD / 10000;
 
     function setUp() public override {
+        emit log_named_uint("block", block.number);
+        emit log_named_uint("timestamp", block.timestamp);
```
hexonaut, line 100: I'm assuming these were left in by mistake?

src/D3MHub.sol
gbalabasquer: It is being used in the `require`.

src/D3MHub.sol
gbalabasquer: Fixed!

src/D3MHub.sol
hexonaut: Sure, but it's just a bit weird to pass something in just to make sure it's the same as another value. I guess it saves gas..

Could maybe think about just removing that check. I'm not sure if it helps us all that much.

src/tests/integration/D3MCompound.t.sol
gbalabasquer: We left them on purpose, so if by any chance we got a failing test we could have the data available to reproduce. However it is true that the normal behavior of `forge` script makes it a bit annoying as always is being printed.
Not sure if we still want to keep it though.

src/D3MHub.sol
gbalabasquer: @talbaneth @hexonaut let me know what you think about it now. I feel I got a much better symmetry just with a bit extra of gas usage.

src/D3MHub.sol
gbalabasquer: Well the problem is otherwise the code is conceptually wrong. We would be comparing ilk debt ceiling against one `urn`'s debt. That's was the main reason to actually use `Art` in the first place, or `art` as now but with the condition that equals to `Art`.

gbalabasquer: It is a bit annoying from the code simplicity perspective, I agree, but this way leaves me with a better mental peace.

src/D3MHub.sol
talbaneth: I like it, I think you can even make the order of the conditions the same as in wind.

gbalabasquer: This guy Tal is hard to get happy eh! lol joking :)
Change done!

talbaneth: lol, perfect now :)

src/tests/D3MHub.t.sol
gbalabasquer: Added the test and did some minor renaming for the other two full amounts ones.

src/tests/D3MHub.t.sol
gbalabasquer: I'm not really opposed to do so, but shouldn't we really do that everywhere.
This might be more interesting to do with Certora probably.

talbaneth: Yeah, it's a bit ugly to add it everywhere.
If we cover it with Certora (or even just try it locally) it's also fine from my side.

talbaneth: I tried it locally for all these types of tests, no failures.

talbaneth (approved):
LGTM

src/D3MHub.sol
hexonaut: Ok, I don't feel strongly. You can leave as-is.

hexonaut (approved):
LGTM

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
-        // - dai debt tracked in vat (CDP or free)
-        uint256 amount = _min(
-                            _min(
-                                supplyReduction,
-                                availableAssets
-                            ),
-                            daiDebt
-                        );
-        require(amount <= MAXINT256, "D3MHub/overflow");
+    function _unwind(bytes32 ilk, ID3MPool _pool, uint256 amount) internal {
```
kmbarry1, line 231: nit (here and elsewhere): why are some arguments prefixed with underscores, and some are not? The meaning of the underscore is unclear to me.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
-        uint256 fees = 0;
-        if (assetBalance > daiDebt) {
-            unchecked {
-                fees = assetBalance - daiDebt;
-            }
+    function _unwind(bytes32 ilk, ID3MPool _pool, address urn) internal {
+        uint256 amount = _pool.maxWithdraw();
+        if (amount > 0) {
+            uint256 toSlip = _min(vat.gem(ilk, urn), amount);
+            require(toSlip <= MAXINT256, "D3MHub/overflow");
```
kmbarry1, line 245: Since the cast below is to a negative `int256`, couldn't this check technically allow up to `MAXINT256 + 1`? Since `int256` goes from `-2^255` up to `2^255 - 1`.

gbalabasquer: We decided to keep it simpler, but also with solc 0.8 and with safe math inbuilt, `-1 * (MININT256)` reverts.
In the prev solidity versions without safe math it did overflow twice without any problem and leaving the original number.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
+            );
+            require(fixInk <= MAXINT256, "D3MHub/overflow");
+            vat.slip(ilk, address(_pool), int256(fixInk)); // Generate extra collateral
+            vat.frob(ilk, address(_pool), address(_pool), address(this), int256(fixInk), 0); // Lock it
+            ink += fixInk;
+            emit Fees(ilk, fixInk);
+        }
+        if (art < ink) { // If there was permissionless DAI paid or fees added as collateral
+            address _vow = vow;
+            uint256 fixArt = ink - art; // Amount of fees + permissionless DAI paid we will now transform to debt
```
iamchrissmith, line 275: can be `unchecked`

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
+    function _normal(bytes32 ilk, ID3MPool _pool, uint256 Art, uint256 lineWad) internal {
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(_pool));
+        require(art == Art, "D3MHub/more-than-one-urn");
+        uint256 currentAssets = _pool.assetBalance(); // Should return DAI owned by D3MPool
+        uint256 maxWithdraw = _pool.maxWithdraw();
+        if (currentAssets > ink) { // If fees were generated
+            uint256 fixInk = _min(
+                currentAssets - ink, // fees generated
+                ink < lineWad // if previously CDP was under debt ceiling
+                    ? (lineWad - ink) + maxWithdraw // up to gap to reach debt ceiling + maxWithdraw
```
iamchrissmith, line 264: if we have set `Line < line` then we could end up in a state where we add `ink` above the Global Debt Ceiling and cannot unwind enough at the end.  We probably want to have a check like `debt > Line ? _divup(debt - Line, RAY) : 0 // global debt ceiling exceeded` here too so `lineWad` becomes either the `Line` or the `line` depending on which is less.

iamchrissmith: we might even be able to move this to `exec` before normal and do something like:

```solidity
uint256 Line = vat.Line();
uint256 lineWad = min(line, Line) / RAY;
_normal(
  ilk,
  _pool,
  Art,
  lineWad
);
```

gbalabasquer: I think this is a different thing. Any `ilk` can go above local debt ceiling or global debt ceiling due fees generation. There is not a restriction for this.
In this case we added this restriction just a safety precaution that something is going wrong with the fee calculation (as we are outsourcing it to a third party).
I think in this case, using the local debt ceiling is more than enough for this purpose as otherwise it gets more complex.
Also note that for knowing if you are passing the general debt ceiling you actually need to calculate the gap between `debt` and `Line` minus what you already have in use.

gbalabasquer: I mean you can't do: `min(line, Line)` as `Line` is not exclusive for this CDP as `line` is.

src/D3MHub.sol
kmbarry1: Okay, sounds good...it doesn't really hurt anything, we'll never hit the edge case where that extra 1 unit of value matters, anyway. Probably the only effect is on people like me who are obsessed with tiny details.

src/D3MHub.sol
kmbarry1: Apologies for necromancing, but is there any chance we would ever possibly have two D3Ms for the same ilk going at once, e.g. as part of a code migration? I figure probably not, worst case you just deploy a new `ilk`, but wanted to float the idea out there.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
+            require(toSlip <= MAXINT256, "D3MHub/overflow");
+            vat.slip(ilk, urn, -int256(toSlip));
 
-            if (amount + fees > availableAssets) {
-                unchecked {
-                    fees = availableAssets - amount;
-                }
-            }
+            _pool.withdraw(amount);
+            daiJoin.join(address(this), amount);
```
kmbarry1, line 249: Just to confirm my understanding here...we intentionally leave `ink` in the pool's urn b/c there is no reason to `grab` and `slip` it away?

src/D3MHub.sol
```diff
@@ -395,126 +390,44 @@ contract D3MHub {
...
 
-        _pool.preDebtChange("exec");
-        uint256 currentAssets = _pool.assetBalance();
+        _pool.preDebtChange();
 
         if (vat.live() == 0) {
             // MCD caged
-            require(end.debt() == 0, "D3MHub/end-debt-already-set");
+            EndLike _end = end;
+            require(_end.debt() == 0, "D3MHub/end-debt-already-set");
```
kmbarry1, line 407: We should document the consequence for the system, either here in comments or the README, if we fail to call `exec` before `End.debt` gets set.

IIUC, that consequence is a haircut to DAI holders in terms of disbursed collateral based on the fraction of the total debt the D3M represents. Although I think the actual market dynamics can be much less straightforward, since only D3M DAI that has been borrowed in the lending market is "real" and borrowers have an incentive to pay up base on AAVE collateralization to obtain DAI and unwind. Take the example of the only collateral is a D3M with N DAI deposited in AAVE, one borrower who has borrowed M < N DAI against ETH, and then sold that M DAI to someone else who is holding it. If the system is caged and `End.thaw()` is called without a successful `skim` of the D3M urn...strictly speaking, the DAI holder gets nothing from the system. BUT, if the AAVE borrower wants his full collateral back, and AAVE's min health factor is H > 1, he's willing to pay $H / DAI to the DAI holder, so the DAI holder actually is more than made whole. In fact if AAVE has a live DAI oracle...the price going up affects the poor borrower's collateralization ratio, so he might have to pay more than H or even be liquidated...

Going further: neither `End.skim()` nor `_unwind()` actually changes `Vat.debt` AFAICT, so I am not convinced we even need this check here! Unless `Pool.postDebtChange` might modify `Vat.debt`...

src/D3MHub.sol
```diff
@@ -395,126 +390,44 @@ contract D3MHub {
...
             require(ilks[ilk].culled == 0, "D3MHub/module-has-to-be-unculled-first");
+            _end.skim(ilk, address(_pool));
```
kmbarry1, line 409: Have we thought through the cases where someone else calls `End.skim` for the pool permissionlessly? At a glance, it should be fine if `culled == 0` b/c `skim` is idempotent, but it could be a problem if `culled == 1`.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
-            diff = ink - art;
-            require(diff <= MAXINT256, "D3MHub/overflow");
-            vat.suck(_vow, _vow, diff * RAY); // This needs to be done to make sure we can deduct sin[vow] and vice in the next call
-            vat.grab(ilk, _pool, _pool, _vow, 0, int256(diff));
+        if (toUnwind > 0) {
+            toUnwind = _min(
+                toUnwind,
+                maxWithdraw
+            );
+            require(toUnwind <= MAXINT256, "D3MHub/overflow");
```
godsflaw, line 314: This can get stuck in the extreme edge case where total `vat.debt` is in excess of `MAXINT256`, the global `Line` gets set to `0`, and `maxWithdraw` is reported in excess of `MAXINT256` by an integration.

More generally, for whatever reason, in a case where both `toUnwind` and `maxWithdraw` are in excess of `MAXINT256` we can prevent revert by simply adding `MAXINT256` to the `_min()` condition.  This would prevent all cases where this contract is bricked because of the `require()`.

NOTE: If others agree with this change, we may want to consider a similar change in the `_wind()` condition below.
```suggestion
            toUnwind = _min(
                _min(
                    toUnwind,
                    maxWithdraw
                ),
                MAXINT256 
            );
```

godsflaw (commented):
Just noticed that in `cull()` the call gets blocked if either `ink` or `art` are over `MAXINT256` which could potentially be a valid state in the extreme.  We could also change this to take the `_min()` of each value and the `MAXINT256` so that we can at least partially `grab()`.  This would require one more `vat.urn()` check after to ensure we collected all of the `ink` and `art` before setting `ilks[ilk].culled` to 1 though.

Not sure what the implications of a partial `cull()` and then a call to `exec()` would be though, so that would require some investigation.

kmbarry1:
> Just noticed that in `cull()` the call gets blocked if either `ink` or `art` are over `MAXINT256` which could potentially be a valid state in the extreme. We could also change this to take the `_min()` of each value and the `MAXINT256` so that we can at least partially `grab()`. This would require one more `vat.urn()` check after to ensure we collected all of the `ink` and `art` before setting `ilks[ilk].culled` to 1 though.
> 
> Not sure what the implications of a partial `cull()` and then a call to `exec()` would be though, so that would require some investigation.

Fwiw I think we can safely assume actual orders of magnitude will not get that high in practice. Other things will break first. In particular since `rate` is rad-unity for the D3M, if `art` for a D3M gets within 10^27 of the maximum `uint256` (which will be much less than `MAXINT256`), `Vat.debt` would overflow--`frob` will not allow that to occur.

kmbarry1:
I know we discussed this (or else I'm fabricating the memory), but what was the resolution/decision/reasoning around `cull` followed by `uncull` when `art < ink` initially? B/c in that case `uncull` recreates more debt than it should, which is unfair to DAI holders. Arguably the hub should cache the `culled` `art` value so `uncull` can work correctly.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
+        emit Unwind(ilk, amount);
+    }
 
-        // Determine the amount of fees to bring back
-        uint256 fees = 0;
-        if (assetBalance > daiDebt) {
-            unchecked {
-                fees = assetBalance - daiDebt;
-            }
+    function _unwind(bytes32 ilk, ID3MPool _pool, address urn) internal {
```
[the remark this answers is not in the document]
kmbarry1: In fact arguably some different verb from "unwind" should be used here, and "(un)wind" should be reserved for functions that `frob`. Maybe this should be `_recover` or `_reconcile` or something like that.

src/D3MHub.sol
kmbarry1, line 314: This check should be moved inside of `_unwind`.

src/D3MHub.sol
```diff
@@ -221,121 +219,118 @@ contract D3MHub {
...
+                                _min(
+                                    lineWad - art, // amount to reach ilk debt ceiling
+                                    (Line - debt) / RAY  // amount to reach global debt ceiling
+                                ),
+                                targetAssets - currentAssets // plan targetAssets
+                            ),
+                            _pool.maxDeposit() // restricts winding if the pool has a max deposit
+                        );
+            }
+            require(art + toWind <= MAXINT256, "D3MHub/wind-overflow");
```
kmbarry1, line 332: Should document `art <= MAXINT256` as an intended invariant of valid D3M implementations in the README or elsewhere.

src/D3MHub.sol
talbaneth: @kmbarry1 can you explain why you expect `ink` to be positive?
This version of `_unwind` (with the address as 3rd parameter) can only be called if the `ilk` is culled (in which case `ink` was grabbed) or after skimming (in which case `ink` is 0 since `ink == art`).

src/D3MHub.sol
gbalabasquer: I couldn't find the usage, as migration should always be able to happen atomically. And in the worst case, as you said, we can deploy with a new `ilk`. It felt much better to keep that invariant in the code.

src/D3MHub.sol
gbalabasquer: The thing with `_pool` is that otherwise it conflicts with the `pool` getter.

src/D3MHub.sol
gbalabasquer: Yeah, the only way to take advantage of that extra wei is using unchecked code, which probably will end looking worse.

src/D3MHub.sol
gbalabasquer: The only exception is when there is permissionless DAI repaid right after ES. Otherwise yeah, I'm a bit lost the case Kurt might be asking.

src/D3MHub.sol
gbalabasquer: So there are two ways to redeem this after shutdown:

1) Just stop `exec` on shutdown and share `aDai` as any other collateral to DAI holders. We could in theory do this, it is the simplest and equals to any other collateral. It is not great though as requires infinite cyclical DAI redemption as we have for example with DAI LPs.

2) Try to recover the most DAI we can, this way we reduce `aDai` shares and the DAI recovered is going to surplus, which is burnt after shutdown, reducing `vat.debt`, that's why we can only do this process until `end.thaw()` is executed.

I just glanced the example but I think I replied the main question, if it is not the case will go deeply on it.

src/D3MHub.sol
gbalabasquer: We need to be careful, and think it better, as the rebalance counts to unwinding the full amount, if we do partial unwinding we can break the rule there.
Before this case wasn't actually possible to reach that value, as winding was always limited to `Art + toWind <= maxint256`, so we couldn't end with a CDP higher than that number.
However now with the rebalance that might happen. What probably is better is limiting the rebalance also to end up `<= maxint256`. Otherwise thinking that there will be that amount of DAI to redeem is because all is already broken.

src/D3MHub.sol
gbalabasquer: > Just noticed that in `cull()` the call gets blocked if either `ink` or `art` are over `MAXINT256` which could potentially be a valid state in the extreme. We could also change this to take the `_min()` of each value and the `MAXINT256` so that we can at least partially `grab()`. This would require one more `vat.urn()` check after to ensure we collected all of the `ink` and `art` before setting `ilks[ilk].culled` to 1 though.
> 
> Not sure what the implications of a partial `cull()` and then a call to `exec()` would be though, so that would require some investigation.

As Kurt said, I think that would be an extreme case that would break other stuff break, and to be honest if it is `cull` breaking, that is something governance can fix. Partial grabbing can bring more unintended consequences.

gbalabasquer:
> Just noticed that in `cull()` the call gets blocked if either `ink` or `art` are over `MAXINT256` which could potentially be a valid state in the extreme. We could also change this to take the `_min()` of each value and the `MAXINT256` so that we can at least partially `grab()`. This would require one more `vat.urn()` check after to ensure we collected all of the `ink` and `art` before setting `ilks[ilk].culled` to 1 though.
> 
> Not sure what the implications of a partial `cull()` and then a call to `exec()` would be though, so that would require some investigation.

As Kurt said, I think that would be an extreme case that would break other stuff break, and to be honest if it is `cull` breaking, that is something governance can fix. Partial grabbing can bring more unintended consequences.

gbalabasquer:
> I know we discussed this (or else I'm fabricating the memory), but what was the resolution/decision/reasoning around `cull` followed by `uncull` when `art < ink` initially? B/c in that case `uncull` recreates more debt than it should, which is unfair to DAI holders. Arguably `cull` should cache the `art` value so `uncull` can work correctly.

Does it really? If you `uncull` a higher amount, more DAI is sent to the surplus buffer, which then cancels out when you skim.

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: Well it is from the perspective of the third party pool. You `wind` it and `unwind` it. We could use `push` and `pull` as well.

src/D3MHub.sol
gbalabasquer: Yeah I guess this is subjective, we did this way to keep very symmetric `_unwind` from `_wind`, and the math to check in `_wind` is a bit more complex. That's the main reason to have it extracted out.
We could even argue if maybe these two internal should just be incorporated to `_normal` as they are called just once and shouldn't provoke stack too deep either. Actually thinking seriously about this now.

src/D3MHub.sol
gbalabasquer: Missed Tal response when replying. Basically what he says.

gbalabasquer:
> > Btw this can't happen, as toWind is restricted to leave the CDP always <= maxint256 so it can not ever happen that after an exec the CDP is left above that number.
> 
> Can we test this as an invariant? I found another two cases (both inline) where we get stuck after global settlement if this value is allowed to be above `MAXINT256`. Knowing the urn cannot get into this potential end state after `exec()` would allay my concerns.

Yeah I think there shouldn't be any problem adding it. Please take a note of it so we don't forget but I'll try to add it asap.

src/D3MHub.sol
kmbarry1: ah yeah that makes sense

src/D3MHub.sol
kmbarry1: sg resolving

src/D3MHub.sol
gbalabasquer: Done!

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: I've removed the regular `_wind` and `_unwind` and put the code in the main internal function, which is called `_exec` now. And the full `_unwind` is called `_wipe` now.

src/D3MHub.sol
gbalabasquer: I've removed the regular `_wind` and `_unwind` and put the code in the main internal function, which is called `_exec` now. And the full `_unwind` is called `_wipe` now.

certora/D3MHub.spec
```diff
@@ -0,0 +1,1376 @@
...
+    );
+    assert(
+        (tic > 0 || !active) &&
+        assetsBefore - maxWithdraw >= inkBefore &&
+        inkBefore > lineWad
+            => artAfter == inkBefore, "unwind: error 11"
+    );
+}
+
+rule exec_normal_revert(bytes32 ilk) {
```
iamchrissmith, line 433: probably need to rename `normal` to `_exec`

certora/D3MHub.spec
```diff
@@ -0,0 +1,1376 @@
...
+    exit(e, ilk, usr, wad);
+
+    uint256 vatGemSenderAfter = vat.gem(ilk, e.msg.sender);
+    uint256 poolShareUsrAfter = share.balanceOf(usr);
+
+    assert(vatGemSenderAfter == vatGemSenderBefore - wad, "vatGemSender did not decrease by wad amount");
+    assert(usr != pool => poolShareUsrAfter == poolShareUsrBefore + wad, "poolShareUsr did not increase by wad amount");
+}
+
+rule exit_revert(bytes32 ilk, address usr, uint256 wad) {
```
iamchrissmith, line 1107: does this cover the case where `wad = 0`? We were thinking that we wanted some protection for this function receiving `wad = 0` for these and future pools?

certora/D3MHub.spec
gbalabasquer: Yeah not sure what is the best as `exec_exec` will clash with the other spec which calls `exec(ilk)` twice.
Also it is more a symbolic name, as we don't have `vat_caged` or `ilk_culled` neither.

certora/D3MHub.spec
gbalabasquer: It includes the value of `wad = 0` basically `wad` can be any `uint256` value.
However the whole suite is working under the use of `D3MTestPool`, so a different `pool` might generate a different behaviour. We might want to think on adding a "MaliciousPool" where we can prove how the `hub` reacts.

src/D3MHub.sol
```diff
@@ -600,4 +466,50 @@ contract D3MHub {
         ilks[ilk].culled = 0;
         emit Uncull(ilk, wad);
     }
+
```
iamchrissmith, line 469: [red team comment] we were thinking we might want to add a `require(end.debt() == 0, "end-debt-set")` to the requires in `uncull`. We know it will not matter because we can't call `exec` once debt > 0, but this would prevent the vat balance changes and reduce the amount of context required to understand `uncull`.  Since it is not a hot path the gas cost shouldn't matter.

gbalabasquer: As we discussed, we shouldn't add this restriction as we can still `hub.uncull(ilk)`, `end.cage(ilk)`. `end.skim(ilk)` and make this collateral available for DAI holders redemption even after `end.wait` passes.

src/D3MHub.sol
```diff
@@ -219,170 +219,123 @@ contract D3MHub {
...
+            unchecked {
+                toUnwind = _max(
+                                _max(
+                                    art > lineWad ? art - lineWad : 0, // ilk debt ceiling exceeded
+                                    debt > Line ? _divup(debt - Line, RAY) : 0 // global debt ceiling exceeded
+                                ),
+                                targetAssets < currentAssets ? currentAssets - targetAssets : 0 // plan targetAssets
+                            );
+                if (toUnwind > 0) {
+                    toUnwind = _min(toUnwind, maxWithdraw);
```
talbaneth, line 304:
```suggestion
                toUnwind = _min(toUnwind, maxWithdraw);
                if (toUnwind == 0) {
```
slightly more compact.

src/D3MHub.sol
```diff
@@ -219,170 +219,123 @@ contract D3MHub {
...
+                } else {
```
talbaneth, line 305:
```suggestion
                toUnwind = _min(toUnwind, maxWithdraw);
                if (toUnwind == 0) {
```
slightly more compact

src/D3MHub.sol
```diff
@@ -219,170 +219,123 @@ contract D3MHub {
...
-    }
+            // Determine if it needs to unwind due to:
+            unchecked {
+                toUnwind = _max(
+                                _max(
+                                    art > lineWad ? art - lineWad : 0, // ilk debt ceiling exceeded
+                                    debt > Line ? _divup(debt - Line, RAY) : 0 // global debt ceiling exceeded
+                                ),
+                                targetAssets < currentAssets ? currentAssets - targetAssets : 0 // plan targetAssets
+                            );
```
talbaneth, line 302:
```suggestion
                           );
```
indentation fix

src/D3MHub.sol
```diff
@@ -219,170 +219,123 @@ contract D3MHub {
...
+        // Determine if fees were generated and try to account them (or the most that it is possible)
+        if (currentAssets > ink) {
+            uint256 fixInk = _min(
+                _min(
+                    currentAssets - ink, // fees generated
+                    ink < lineWad // if previously CDP was under debt ceiling
+                        ? (lineWad - ink) + maxWithdraw // up to gap to reach debt ceiling + maxWithdraw
+                        : maxWithdraw // up to maxWithdraw
+                ),
+                SAFEMAX + art - ink //  ensures that fixArt * RAY (rate) will be <= MAXINT256 (in vat.grab)
```
talbaneth, line 258: Not sure I understand the comment.
Isn't `fixArt * RAY <= MAXINT256` `because `fixArt = ink - art` and `ink < SAFEMAX`?
Maybe the comment should be about why the min calculation includes `SAFEMAX + art - ink`.

src/D3MHub.sol
gbalabasquer: Ah yes, this doesn't work if `maxWithdraw` equals to 0. Which is why the test failed.

src/D3MHub.sol
gbalabasquer: So we need to make sure `fixArt * RAY` is less than `MAXINT256` as otherwise in `vat.grab`, this line `int dtab = mul(ilk.rate, dart);` will fail.
It's true that `ink` is `<= SAFEMAX`, however `ink` can previously be increased by `currentAssets`. So the whole idea of including this part in the `min` is to make sure that the end resulted `fixArt`, which will be ` >= fixInk` doesn't overflow.
Not sure how to make the comment clearer though.

src/D3MHub.sol
gbalabasquer: Actually I always have issues with these type of cases as if I remove a space it breaks the 4 spaces of the tab.

src/D3MHub.sol
gbalabasquer: This thing should be addressed now.

src/D3MHub.sol
gbalabasquer: I think this issue was totally solved with latest functionality, so can probably be resolved.

src/D3MHub.sol
kmbarry1: Okay think we're good here.

src/D3MHub.sol
```diff
@@ -219,170 +219,123 @@ contract D3MHub {
...
+        if (amount > 0) {
+            _pool.withdraw(amount);
+            daiJoin.join(address(this), amount);
+            vat.move(address(this), vow, amount * RAY);
 
-        emit Wind(ilk, amount);
+            uint256 toSlip = _min(vat.gem(ilk, urn), amount);
+            // amount bounds toSlip and amount * RAY bounds amount to be much less than MAXINT256
+            vat.slip(ilk, urn, -int256(toSlip));
+            emit Unwind(ilk, amount);
```
kmbarry1, line 234: Do we want a `Wipe` event here instead of `Unwind`? Might be good to avoid ambiguity.

src/D3MHub.sol
kmbarry1: I think I answered my own question about the need to do this...since we transfer dai to the Vow, someone could then call `heal` and change `debt`.

IIUC, case 1. can still happen even if we fail to execute this flow? Think I missed that in my analysis.

src/D3MHub.sol
gbalabasquer: Hmmm my personal opinion is `Unwind` is just fine, as they can only happen in different stages of the contract (meaning that there is certainty on which one was used), and also as in fact they are doing the exact same thing.
It is not like within `wipe` we will clean the whole CDP for sure, we are just unwinding with a different way to calculate the amount.
But will leave this open to the rest to give their opinion.

src/D3MHub.sol
gbalabasquer: Yeah case 1 can still happen if nobody executes `exec` during the `end.wait` time.
Anyone can do `end.skim(ilk)`, that doesn't change `vat.debt`, just `vat.vice` and `vat.sin(vow)` (which are not used in any `End` distribution accounting). The only result should be that now the `end` has `aDai` collateral to share the `DAI` holders.

The advantage of calling `exec` before `end.thaw` is actually trying to unwind collateral before settlement, in that case it will in fact actually affect `vat.debt` for collateral distribution. There will be less DAI in circulation (`vat.debt`) but also less (or none) `aDai` to give to the DAI holders.
In case we do not call it we will have more `aDai` with more DAI in circulation, which in fact if you cycle in an infinite way from DAI -> aDai -> DAI -> ... you should end up with the same amount of the other collaterals in both cases (assuming there is liquidity to withdraw DAI from aDai).

src/D3MHub.sol
gbalabasquer: @kmbarry1 let me know if we can resolve this one.

src/D3MHub.sol
gbalabasquer: This was a bit changed. Now the `require` is `art + toWind <= MAXINT256 / RAY`.
Basically the whole logic is now designed now to work with `ink` and `art` being less or equal to that value, so when we have `int256 dtab = dart * rate (RAY)` in `vat.frob` or `vat.grab` it can not overflow.
We also do these explicit `require`s: `ink >= art` and `ink <= MAXINT256 / RAY` at the beginning of the normal path of `exec` and there is a Certora assert in the spec that checks that `ink` can not end up higher than that number after `exec`.

src/D3MHub.sol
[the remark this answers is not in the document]
gbalabasquer: @kmbarry1 let me know if we can resolve this one.

src/D3MHub.sol
iamchrissmith: I like `Unwind` as (like gonza pointed out) we might not be able to wipe completely

kmbarry1: I was just thinking it would be good to clear which code path had been triggered...but don't feel too strongly.

src/D3MHub.sol
kmbarry1: Yeah, believe I was thinking about the case Gonza highlights. Worst case I guess some lending pool shares get stuck in the D3M? This isn't a normal Vault where the owner can claim their collateral with a `free` call.

gbalabasquer: Yeah the only case where `ink` might be higher than `art` is if there was a permissionless repayment of DAI right before ES (not after as I said in the prev comment). As there wasn't time to recall `exec` to fix it. In that case that difference of `ink` will be lost.

Then the other thing that might be lost is the different between the `ink` and the real amount of assets accrued in the pool. That won't be updated after ES in `exec`.
So if before `end.wait` finishes, there is liquidity, all of it will be recovered, but otherwise only `ink` amount will be shared to DAI holders.

PR #87, opened 2022-09-21
More precise pool shares

gbalabasquer:


src/D3MHub.sol
```diff
@@ -283,8 +284,10 @@ contract D3MHub {
         uint256 toUnwind;
         uint256 toWind;
 
-        // Determine if it needs to fully unwind due to D3M ilk being caged (but not culled) or plan is not active
-        if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active()) {
+        // Determine if it needs to fully unwind due to D3M ilk being caged (but not culled), plan is not active or something
+        // wrong is going with the third party and we are entering in the ilegal situation of having less assets than registered
+        // It's adding up `WAD` due possible rounding errors
+        if (ilks[ilk].tic != 0 || !ilks[ilk].plan.active() || currentAssets + WAD < ink) {
```
gbalabasquer, line 290: Instead of `WAD` I think we could just do 1 in the precision diff between DAI and the token. For Aave 1 wei should be enough, for Compound I think the margin should be 10^10. But doing that different per `ilk` would require ro add that number somehow in the `pool` or `plan` and I don't think it is probably worth it. 1 WAD is not much difference to really care.

src/pools/D3MAavePool.sol
```diff
@@ -173,8 +181,10 @@ contract D3MAavePool is ID3MPool {
         require(dai.balanceOf(msg.sender) == prevDai + wad, "D3MAavePool/incorrect-dai-balance-received");
     }
 
-    function transfer(address dst, uint256 wad) external override onlyHub {
-        require(adai.transfer(dst, wad), "D3MAavePool/transfer-failed");
+    function exit(address dst, uint256 wad) external override onlyHub {
+        uint256 amt = wad * assetBalance() / (D3mHubLike(hub).end().Art(ilk) - exited);
```
talbaneth, line 185: Logic looks good.

If we care much about gas here we can:
* Consider keeping `end` in storage (fileable).
* Caching `exited` in the beginning and doing something like:  `exited = _exited +  wad`

src/pools/ID3MPool.sol
```diff
@@ -51,7 +51,7 @@ interface ID3MPool {
         @param dst address that should receive the redeemable tokens
         @param wad amount in Dai terms that we want to withdraw
     */
-    function transfer(address dst, uint256 wad) external;
+    function exit(address dst, uint256 wad) external;
```
talbaneth, line 54: Maybe we should update the documentation a bit.
I guess at least the @notice part should be something like "Exit proportional amount of shares".

src/tests/stubs/D3MTestPool.sol
```diff
@@ -144,7 +149,7 @@ contract D3MTestPool is ID3MPool {
     }
 
     function maxWithdraw() external view override returns (uint256) {
-        return TokenLike(dai).balanceOf(share);
+        return _min(TokenLike(dai).balanceOf(share), assetBalance());
```
talbaneth, line 152: Can you explain this change please.

src/tests/pools/D3MCompoundPool.t.sol
```diff
@@ -218,35 +217,39 @@ contract D3MCompoundPoolTest is D3MPoolBaseTest {
         assertEq(pool.redeemable(), address(cDai));
     }
 
-    function test_transfer_cdai() public {
-        pool.file("hub", address(this));
+    function test_exit_cdai() public {
```
talbaneth, line 220: We might want a test that does 2 exits with an elapsing blocks (time in aave) period between them.
We want to make sure all dai and cdai were handed out.

src/pools/D3MAavePool.sol
gbalabasquer: - I started with a fileable `end` but then when I realized it could come from the `hub` change to the actual one. I know is more gas intensive but at the same time will save us complication for initial set up and in case of `End` migration.
Anyway not that strong feeling if you prefer the other way.
- Done

src/tests/stubs/D3MTestPool.sol
gbalabasquer: I had a test failing due this. The correct behaviour of `maxWithdraw` should be the minimum between the shares of DAI owned by our pool contract in the third party pool and the total available DAI in that moment in the whole pool. So basically now is doing what the real pools do.

gbalabasquer: This was the test failing: `test_wind_limited_by_pool_loss` and it makes sense as share decreased from our balance.
Which btw I'm now renaming to: `test_unwind_due_to_by_pool_loss`

src/pools/D3MAavePool.sol
talbaneth: Ok, I'm also fine with the way `end` is fetched from the hub.
Resolving.

src/tests/stubs/D3MTestPool.sol
talbaneth: perfect

src/pools/D3MAavePool.sol
```diff
@@ -173,8 +181,11 @@ contract D3MAavePool is ID3MPool {
         require(dai.balanceOf(msg.sender) == prevDai + wad, "D3MAavePool/incorrect-dai-balance-received");
     }
 
-    function transfer(address dst, uint256 wad) external override onlyHub {
-        require(adai.transfer(dst, wad), "D3MAavePool/transfer-failed");
+    function exit(address dst, uint256 wad) external override onlyHub {
+        uint256 exited_ = exited;
+        exited = exited_ + wad;
+        uint256 amt = wad * assetBalance() / (D3mHubLike(hub).end().Art(ilk) - exited_);
```
kmbarry1, line 187: There's an implicit invariant here, right? That `end().Art(ilk)` needs to be at least as much as the exitable `ink`. Couldn't that be violated in some repayment scenarios?

kmbarry1: Not suggesting we need to block on it here, just calibrating my understanding for unit testing.

src/tests/pools/D3MCompoundPool.t.sol
gbalabasquer: Added new tests.

src/D3MHub.sol
gbalabasquer: Other thing that realized and discussed with @talbaneth, adding up `WAD` means that a malicious value of `currentAssets > maxinuint256 - WAD` will make a potential unwinding to stop.
We agreed that caging or making the plan inactive (which won't require gov delay) it is probably more than enough protection for it (as the conditional will jump before than evaluating it). cc @kmbarry1

talbaneth (approved):
LGTM

src/D3MHub.sol
kmbarry1: Good edge case to call out but I doubt it would be hit in practice, and if someone is able to manipulate the `currentAssets` value (or if it has a bug) action would be necessary anyway. Don't think the side effects here are terribly severe, either.

gbalabasquer: Yeah seems we are all align regarding this. Thanks for your feedback!

src/tests/integration/D3MAave.t.sol
```diff
@@ -949,16 +945,68 @@ contract D3MAaveTest is DSSTest {
...
+
+        uint256 totalArt = end.Art(ilk);
+
+        // Simulate DAI holder gets some gems from GS
+        vm.prank(address(end));
+        vat.flux(ilk, address(end), address(this), totalArt);
+
+        assertEq(adai.balanceOf(address(this)), 0);
+
+        // User can exit and get the cDAI
```
iamchrissmith, line 983:
```suggestion
        // User can exit and get the aDAI
```

src/pools/D3MAavePool.sol
```diff
@@ -86,7 +93,8 @@ contract D3MAavePool is ID3MPool {
     event File(bytes32 indexed what, address data);
     event Collect(address indexed king, address indexed gift, uint256 amt);
 
-    constructor(address hub_, address dai_, address pool_) {
+    constructor(bytes32 ilk_, address hub_, address dai_, address pool_) {
+        ilk = ilk_;
```
iamchrissmith, line 97: rather than passing this in for each pool in the constructor (kind of hard to enforce in our current interface pattern), how would you feel about adding it to the `exit` args?

src/tests/integration/D3MCompound.t.sol
```diff
@@ -989,17 +985,72 @@ contract D3MCompoundTest is DSSTest {
...
+        // Simulate DAI holder gets some gems from GS
+        vm.prank(address(end));
+        vat.flux(ilk, address(end), address(this), totalArt);
+
+        uint256 initialCDaiBalance = cDai.balanceOf(address(d3mCompoundPool));
+
+        assertEq(cDai.balanceOf(address(this)), 0);
+
+        // User can exit and get the cDAI
+        uint256 expectedCDai = 25 * 1e18 * initialCDaiBalance / totalArt;
```
iamchrissmith, line 1026: why switch to `1e18` from `WAD`?

src/tests/pools/D3MAavePool.t.sol
```diff
@@ -249,28 +247,30 @@ contract D3MAavePoolTest is D3MPoolBaseTest {
...
         uint256 tokens = adai.totalSupply();
         adai.transfer(d3mTestPool, tokens);
         assertEq(adai.balanceOf(address(this)), 0);
         assertEq(adai.balanceOf(d3mTestPool), tokens);
 
-        D3MAavePool(d3mTestPool).file("hub", address(this));
-        D3MAavePool(d3mTestPool).transfer(address(this), tokens);
+        end.setArt(tokens);
+        vm.prank(hub);
+        D3MAavePool(d3mTestPool).exit(address(this), tokens);
```
iamchrissmith, line 258: Am I understanding this correctly? `vm.prank` do not change the value of `address(this)` it only changes `msg.sender`?

src/tests/pools/D3MAavePool.t.sol
```diff
@@ -249,28 +247,30 @@ contract D3MAavePoolTest is D3MPoolBaseTest {
         assertEq(D3MAavePool(d3mTestPool).redeemable(), address(adai));
     }
 
-    function test_transfer_adai() public {
+    function test_exit_adai() public {
```
iamchrissmith, line 250: since we now don't expect this to be called except during ES, maybe we can remove this test? I think it is a hold over from when `transfer` was for upgrading too

src/tests/pools/D3MCompoundPool.t.sol
iamchrissmith, line 220: same comment as on Aave, I think we can remove this test and just leave the `caged` one

Makefile
```diff
@@ -1,6 +1,5 @@
...
-certora-hub  :; PATH=~/.solc-select/artifacts:${PATH} certoraRun --solc_map D3MHub=solc-0.8.14,Vat=solc-0.5.12,DaiJoin=solc-0.5.12,Dai=solc-0.5.12,End=solc-0.5.12,D3MTestPlan=solc-0.8.14,D3MTestPool=solc-0.8.14,D3MTestGem=solc-0.8.14 --rule_sanity basic src/D3MHub.sol certora/dss/Vat.sol certora/dss/DaiJoin.sol certora/dss/Dai.sol certora/dss/End.sol src/tests/stubs/D3MTestPlan.sol src/tests/stubs/D3MTestPool.sol src/tests/stubs/D3MTestGem.sol --link D3MHub:vat=Vat D3MHub:daiJoin=DaiJoin D3MHub:end=End DaiJoin:vat=Vat DaiJoin:dai=Dai End:vat=Vat D3MTestPlan:dai=Dai D3MTestPool:hub=D3MHub D3MTestPool:vat=Vat D3MTestPool:dai=Dai D3MTestPool:share=D3MTestGem --verify D3MHub:certora/D3MHub.spec --settings -mediumTimeout=180 --staging$(if $(short), --short_output,)$(if $(rule), --rule $(rule),)$(if $(multi), --multi_assert_check,)
+all			:; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.8.14 build
+clean		:; dapp clean
+test		:; ./test.sh match="$(match)" block="$(block)" match-test="$(match-test)" match-contract="$(match-contract)"
+test-forge	:; make test
```
iamchrissmith, line 4: will this pass along arguments if passed?

src/tests/pools/D3MAavePool.t.sol
gbalabasquer: Right, that's how I understand it.

src/tests/integration/D3MCompound.t.sol
gbalabasquer: That's how it was already in the cDai tests

src/tests/pools/D3MAavePool.t.sol
gbalabasquer: To be honest I think the one that doesn't make sense is the vat caged one. This is a unit test for the `pool` and the `vat` status has nothing to do with this method.

Makefile
gbalabasquer: I just checked and it does, eventually we should directly remove it though.

src/pools/D3MAavePool.sol
gbalabasquer: I considered this option but eventually felt it was more an implementation detail for these two pools, that's why ended up passing the `ilk` in the constructor. If we consider this is actually something to generalize, then yeah we could move to the function param.

src/tests/pools/D3MAavePool.t.sol
iamchrissmith: actually, you're right.  Lets remove the cage one?

src/tests/integration/D3MCompound.t.sol
iamchrissmith: it looked like it was just a few places in those, so i created this to change to WAD for consistency: https://github.com/makerdao/dss-direct-deposit/pull/88

src/tests/pools/D3MCompoundPool.t.sol
gbalabasquer: Removed the `cage` on as discussed in the other comment.

src/tests/integration/D3MCompound.t.sol
gbalabasquer: Merged!

iamchrissmith (approved):
looks good to me. before you merge, we should check with @godsflaw and @naszam to see if they would prefer to review these changes here or in the main branch the red team is reviewing

godsflaw:
You guys going to rug me and merge this?  I will have to start over anyway (if I'm following the strict format of the matrix), so I'm okay with that.  It already happened to me once.  I would rather review this change in place than wonder if it invalidates my review.

kmbarry1:
> You guys going to rug me and merge this? I will have to start over anyway (if I'm following the strict format of the matrix), so I'm okay with that. It already happened to me once. I would rather review this change in place than wonder if it invalidates my review.

In any case I've been waiting for this PR to get finalized before writing exhaustive tests...happy to write them against this branch if we want to wait on the merge.

PR #90, opened 2022-10-13
Matrix Tests

godsflaw:
Working through some tests from the matrix: https://hackmd.io/@BD4VcoOwSHauNSOi3I4eMw/SJsUIi8ks


src/tests/D3MHub.t.sol
```diff
@@ -716,6 +830,18 @@ contract D3MHubTest is DSSTest {
...
+    function test_wind_mcd_caged() public {
+        vat.file("Line", vat.debt() + 45 * RAD);
+        vat.file(ilk, "line", 40 * RAD);
+
+        // MCD shuts down
+        end.cage();
+        end.cage(ilk);
+
+        // try to wind the system
+        _windSystemNoop();
```
talbaneth, line 842: Is this really a no-op? I think we are unwinding due to mcd being caged.

src/tests/D3MHub.t.sol
```diff
@@ -770,6 +896,144 @@ contract D3MHubTest is DSSTest {
         assertEq(vat.sin(vow), sinBefore + 40 * RAD);
     }
 
+    function test_wind_pool_caged() public {
```
talbaneth, line 899: Mmm what is the difference between this and the existing `test_unwind_pool_caged`?

So there is already a `test_unwind_*` series of tests, named after the fact that the action we do (not explicilty unwinding) will cause an unwind.

The new `test_wind_*`  series is basically the same, and confused me. We should try to merge them to the existing format.

src/tests/D3MHub.t.sol
```diff
@@ -1076,6 +1382,108 @@ contract D3MHubTest is DSSTest {
...
+        assertEq(vat.dai(address(d3mHub)), 0);
+        // Sin only increases by 40 WAD since 10 was covered previously
+        assertEq(vat.sin(vow), sinBefore + 40 * RAD);
+        assertEq(vat.dai(vow), vowDaiBefore);
+        (, , , uint256 culled, ) = d3mHub.ilks(ilk);
+        assertEq(culled, 1);
+
+        d3mTestPlan.file("bar", 10);
+        d3mTestPlan.file("targetAssets", 50 * WAD);
+        vat.file(ilk, "line", 40 * RAD);
```
talbaneth, line 1425: Mmm seems this and the test afterwards replicate `test_cull_debt_paid_back` but with 3 additionals lines each that actually make no difference since we are culled and will unwind as much as possible.
Suggest removing the 2 extra tests for the sake of readability.

src/tests/D3MHub.t.sol
gbalabasquer: Yeah I think there is a misleading naming here. I think you could also use Foundry's power to check the events, that would allow you to prove better that the winding path is not being taken (what I see is you really trying to test).

gbalabasquer: I also think the internal function makes sense to make a pre state, like in every (or almost) test we need to wind the system to something, then we have `_windSystem()`. However after that I'd expect special conditions and checks happening in the same test itself.

src/tests/D3MHub.t.sol
gbalabasquer: Yeah these two seems to be equivalent. I guess the naming is subjective, but I prefer the naming that reflects what actually happens than the what would you might expect to happen but are testing is not the case.

gbalabasquer: So in this case, pool is caged, so what you are testing is that does NOT wind more and actually unwinds.

gbalabasquer:
I have not gone deeply through the tests, but I feel there is certain stuff which is much easier and it is more powerful to check with Certora. Like, if a pool is cage, make sure we never wind. I think that can be done within a specific FV spec and we would avoid a huge amount of unit tests that also doesn't give you a high confidence you are still going through all the paths.
For example, as Tal mentioned in one of the comments, if the system is culled, it doesn't really matter what happens with the `Line` if you already have a spec that proves that in that condition will always try to unwind the max it can.

PR #91, opened 2022-11-22
Add test_borrow_apy

rockyfour:


iamchrissmith (approved):
lgtm

src/tests/integration/D3MCompound.t.sol
```diff
@@ -69,6 +71,9 @@ interface InterestRateModelLike {
 }
 
 contract D3MCompoundTest is DSSTest {
+    Hevm hevm;
```
gbalabasquer, line 74: We have deprecated `dapp` and `hevm`, you should use directly `vm` (without declaring it) and cheat code is also not necessary.

src/tests/integration/D3MCompound.t.sol
```diff
@@ -300,26 +306,28 @@ contract D3MCompoundTest is DSSTest {
...
-        uint256 amountSupplied = cDai.balanceOfUnderlying(address(d3mCompoundPool));
-        assertTrue(amountSupplied > 0);
-        assertLt(amountSupplied, amountSuppliedInitial);
-        (uint256 ink, uint256 art) = vat.urns(ilk, address(d3mCompoundPool));
-        assertEqRounding(ink, amountSupplied);
-        assertEqRounding(art, amountSupplied);
+        // fast forward 1 year while accruing interest each day
+        for (uint256 i = 1; i <= 365; i++) {
+            hevm.roll(block.number + 7200);
+            cDai.accrueInterest();
```
gbalabasquer, line 324: Why is it needed to be done every day?

src/tests/integration/D3MCompound.t.sol
rockyfour: right, done

src/tests/integration/D3MCompound.t.sol
rockyfour: Not sure, but it seems to improve the accrual precision - when expecting 1.02 I'm getting 1.020146 instead of 1.019803 (when just doing `hevm.roll(block.number + 7200 * 365)`).
It is probably a more realistic case as accrual will be done on any borrow/repay.

gbalabasquer: ok got it

gbalabasquer (approved):
LGTM. Ignore the Certora specs failing, I already reported to them, and it is also happening on `v2`.

PR #92, opened 2022-12-16
Add Aave and Compound Deploy Scripts

hexonaut:
This will add a script to deploy / init the Aave and Compound D3Ms.

On the proposed structure:

I've split into separate scripts / libraries. The scripts allow the repository to connect to the command line with a forked anvil node, etc.  Basically it's a thin wrapper on the libraries. The libraries allow more flexibility in that they can be called inside integration tests or the spell.

To solve the issue of version conflicts I recommend splitting the deploy (`D3MDeploy.sol`) and init (`D3MInit.sol`) stages such that the init only pulls in interfaces either defined inline or pulled in from `dss-interfaces`. To run the library in a script we will add it to `spells-mainnet`. We can then import the standalone init file and just call that.

So here is what it will all look like:

#### Integration Tests

```solidity
D3MDeploy.deployAave(...);
vm.startPrank(pauseProxy);
D3MInit.initAave(...);
vm.stopPrank();
```

#### Anvil Fork / Bash Script

```bash
# An example where the D3M config comes from a local config and the Aave lending pool is dynamically added to the config (Perhaps it was deployed in another script above)
export FOUNDRY_SCRIPT_CONFIG_TEXT=`jq -c ". + { lendingPool: \"$FOUNDRY_EXPORT_LENDING_POOL\" }" < /path/to/myCustomConfig.json`

forge script script/DeployD3M.s.sol:DeployD3M --use solc:0.8.14 --rpc-url $ETH_RPC_URL --sender $ETH_FROM --broadcast

cast rpc anvil_setBalance $MCD_PAUSE_PROXY 0x10000000000000000
cast rpc anvil_impersonateAccount $MCD_PAUSE_PROXY
unset ETH_FROM
forge script script/InitD3M.s.sol:InitD3M --use solc:0.8.14 --rpc-url $ETH_RPC_URL --broadcast --unlocked --sender $MCD_PAUSE_PROXY
cast rpc anvil_stopImpersonatingAccount $MCD_PAUSE_PROXY
```

#### Spell

Deploy:

```bash
# Just use the repository's Aave config
export FOUNDRY_SCRIPT_CONFIG="aave"

forge script script/DeployD3M.s.sol:DeployD3M --use solc:0.8.14 --rpc-url $ETH_RPC_URL --sender $ETH_FROM --broadcast
```

Spell:

```solidity
import "dss-direct-deposit/deploy/D3MInit.sol";

...


// Inside the action
D3MInit.initAave(...);
```


Makefile
```diff
@@ -1,5 +1,4 @@
-all         :; DAPP_BUILD_OPTIMIZE=1 DAPP_BUILD_OPTIMIZE_RUNS=200 dapp --use solc:0.8.14 build
-clean       :; dapp clean
+all         :; forge build --use solc:0.8.14
```
gbalabasquer, line 1:
```suggestion
all         :; FOUNDRY_OPTIMIZER=true FOUNDRY_OPTIMIZER_RUNS=200 forge build --use solc:0.8.14
```
Even if it is the default might make sense to be very explicit.

Makefile
```diff
@@ -1,5 +1,4 @@
...
+clean       :; forge clean
 test        :; ./test.sh solc="0.8.14" match="$(match)" block="$(block)" match-test="$(match-test)" match-contract="$(match-contract)"
```
gbalabasquer, line 3:
```suggestion
test        :; ./test.sh solc="0.8.14" match="$(match)" block="$(block)" match-test="$(match-test)" match-contract="$(match-contract)"
deploy.   :; ./deploy.sh
```

script/input/1/config.json
```diff
@@ -0,0 +1,21 @@
...
+    "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
+    "admin": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
+    "hub": "0x12F36cdEA3A28C35aC8C6Cc71D9265c17C74A27F",
+    "type": "aave",
+    "ilk": "DIRECT-AAVEV2-DAI",
+    "maxLine": 5000000,
+    "gap": 3000000,
+    "ttl": 28800,
+    "tau": 604800,
+    "aave": {
```
gbalabasquer, line 11: I'm a bit confused about this structure of file. The `ilk` and auto-line settings are per each D3M. Why are they in an upper level and then we have `aave` and `compound` sub values?

gbalabasquer: I guess we define the general and then the specifics, but I feel a bit confusing to have it mixed this way. As I'm gonna a deploy aave (for example) my config file won't have anything about compound.

hexonaut: I'm open to alternatives, but I think this makes sense as both Aave and Compound adapters share all the parameters in the root object. They both have debt ceilings and whatnot.

It's possible you are thinking this are parameters that should be defined per-d3m? These are just sensible defaults to get up and running quickly and most real deploys will override whatever they need to via environment variables. There are examples of this in the OP examples above.

In the case of the spell init this will be hard coded in the spells same as always and these values will be unused.

rockyfour: I think it might be cleaner to try and use only the config file parameters and not a mixup of env vars and the config file parameters.
That way there is less chance for mistakes (it's always confusing to use overriding env vars in a script). Also this way the config file is an afterward record of what was used (assuming the changes are committed).

hexonaut: We could use a full json config file as an override, but the structure of having a file-based override is not that clean. There are two points:

1. There may be more than just "official" Aave and Compound deploys. People can deploy forks of these protocols. If/when we add something like Centrifuge D3Ms they will have multiple deploys as well. (5 instances are live now). It's a bit awkward to have the configs here match what's live. Do we keep these synced when parameters changes are made over time after initial deploy/init?
2. If I want to deploy a D3M to a chain fork which is the sole use-case for even calling `init` with a config file then it is much easier to fork the repo as-is and just define the environment variable overrides. Otherwise you need to either add a json file config override to the repository or add file permissions to some random file elsewhere in the system. A common pattern for this will be to have a higher-level deploy script repository which imports `dss-direct-deposit` and other dependencies.

An example structure:

```
dependencies/dss-teleport/...
dependencies/xdomain-dss/...
dependencies/dss-bridge/...
deploy-domain.sh
init-domain.sh
```

`deploy.sh` and `init.sh` would then just be able to change into the dependencies directories of each repository and run each deploy/init script respectively against an anvil fork without having to modify the repositories. As I said feeding in an override config is possible too, but we have to deal with file permissions then.

Overall I would prefer a higher level coordination structure defined in Foundry, and there has been an [issue opened for this](https://github.com/foundry-rs/foundry/issues/3911), but for now the environment variable overrides works quite nicely, but again if you guys feel strongly about going the override route we can find something that works.

hexonaut: Foundry team seems like they are willing to move on this actually. I've added a point about standardizing global config with the `deploy` command: https://github.com/foundry-rs/foundry/issues/3911#issuecomment-1369962540

Let's hold off merging this until we see what that looks like.

gbalabasquer: Honestly my point was more about that looks a bit weird the example file has defined an `ilk` called `DIRECT-AAVEV2-DAI` and it says type `aave` and then we have a compound section.

So I would expect something more like:
```
{
    "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
    "admin": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
    "hub": "0x12F36cdEA3A28C35aC8C6Cc71D9265c17C74A27F",
    "type": "aave",
    "ilk": "DIRECT-AAVEV2-DAI",
    "maxLine": 5000000,
    "gap": 3000000,
    "ttl": 28800,
    "tau": 604800,
    "lendingPool": "0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9",
    "king": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
    "bar": 370
}
```
and
```
{
    "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
    "admin": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
    "hub": "0x12F36cdEA3A28C35aC8C6Cc71D9265c17C74A27F",
    "type": "compound",
    "ilk": "DIRECT-COMP-DAI",
    "maxLine": 5000000,
    "gap": 3000000,
    "ttl": 28800,
    "tau": 604800,
    "cdai": "0x5d3a536E4D6DbD6114cc1Ead35777bAB948E3643",
    "king": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
    "barb": 7535450719
}
```

gbalabasquer: I can agree with Tal, that replacing with env variables is confusing.
IMO what we should leave in the repo is a config example and then have to pass this config file as a flag of the deployment scripts. So user locally defines the file and then pases the path to it when executing.

hexonaut: Ah I see what you're saying. Ok yes I like that structure better. Can split the config file into two examples (Aave and Compound) and pass in full config via cli somehow (it was awkward to do this as you can't read cli args AFAIK).

gbalabasquer: I guess we use the bash script to get the file from the flag and pass the values. At least in the meanwhile we use the bash.

hexonaut: Updated to use only json config files.

script/input/1/aave.json
```diff
@@ -0,0 +1,15 @@
+{
+    "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
+    "admin": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
```
gbalabasquer, line 3: Do we need a key for admin or we can use always the pause proxy from the chainlog?

hexonaut: I think it's helpful to not assume pause proxy here. It's possible someone wants to make a deploy under an EOA for testing purposes or something.

src/deploy/D3MInit.sol
```diff
@@ -0,0 +1,266 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.8.0;
+
+import "dss-interfaces/Interfaces.sol";
```
gbalabasquer, line 19: I'd import only the interfaces we want to use, more thinking this will be used for spells.

src/deploy/D3MInit.sol
```diff
@@ -0,0 +1,266 @@
...
+
+        hub.file(ilk, "pool", d3m.pool);
+        hub.file(ilk, "plan", d3m.plan);
+        hub.file(ilk, "tau", cfg.tau);
+
+        oracle.file("hub", address(hub));
+
+        dss.spotter.file(ilk, "pip", address(oracle));
+        dss.spotter.file(ilk, "mat", 10 ** 27);
+        if (!cfg.existingIlk) {
```
gbalabasquer, line 169: Suggestion: remove this extra config param and use `try {}`

src/deploy/D3MInit.sol
```diff
@@ -0,0 +1,266 @@
...
+            dss.vat.init(ilk);
+            dss.jug.init(ilk);
+        }
+        dss.vat.file(ilk, "line", cfg.gap);
+        dss.vat.file("Line", dss.vat.Line() + cfg.gap);
```
gbalabasquer, line 174: This should take into account if there was already a ilk.line before and discount it.

src/deploy/D3MInit.sol
hexonaut: Personally I'd prefer to be explicit rather than trying to smooth over. Do you feel strongly to switch to try catch behaviour?

src/deploy/D3MInit.sol
hexonaut: done

src/deploy/D3MInit.sol
hexonaut: done

src/deploy/D3MInit.sol
```diff
@@ -16,7 +16,11 @@
 
 pragma solidity >=0.8.0;
 
-import "dss-interfaces/Interfaces.sol";
+import {
```
gbalabasquer, line 19: I don't think this makes a change for the flatten script, it will still load everything.  We should be doing instead:
```
import "dss-interfaces/dss/DssAutoLineAbstract.sol";
import "dss-interfaces/dss/IlkRegistryAbstract.sol";
import "dss-interfaces/ERC/GemAbstract.sol";
```

hexonaut: done

src/deploy/D3MInit.sol
gbalabasquer: I was just thinking in reduce the config file, but you might be right and being explicit could be better.

script/D3MCoreDeploy.s.sol
```diff
@@ -0,0 +1,58 @@
...
+contract D3MCoreDeployScript is Script {
+
+    using stdJson for string;
+    using ScriptTools for string;
+
+    string config;
+    DssInstance dss;
+
+    address admin;
+    D3MCoreInstance d3mCore;
```
gbalabasquer, line 38: Any reason all of this is declared globally? As there is only one function that use it. I'm a bit confused if it is just a personal preference or actually has an intention?

deploy-core.sh
```diff
@@ -0,0 +1,21 @@
...
+    echo "Invalid chainid of $FOUNDRY_ROOT_CHAINID. Please set your forking environment via ETH_RPC_URL or manually by defining FOUNDRY_ROOT_CHAINID (1 or 5)."
+    exit 1;
+}
+
+[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] && echo "Deploying D3M Core on Mainnet"
+[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Deploying D3M Core on Goerli"
+
+rm -f out/contract-exports.env
+export FOUNDRY_ROOT_CHAINID
+forge script script/D3MCoreDeploy.s.sol:D3MCoreDeployScript --use solc:0.8.14 --rpc-url $ETH_RPC_URL --sender $ETH_FROM --broadcast --verify
```
gbalabasquer, line 21:
```suggestion
    FOUNDRY_ROOT_CHAINID="$(cast chain-id)"
}
[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] || [[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] || {
    echo "Invalid chainid of $FOUNDRY_ROOT_CHAINID. Please set your forking environment via ETH_RPC_URL or manually by defining FOUNDRY_ROOT_CHAINID (1 or 5)."
    exit 1;
}

[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] && echo "Deploying D3M Core on Mainnet"
[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Deploying D3M Core on Goerli"

export FOUNDRY_ROOT_CHAINID
rm -f out/contract-exports.env
export FOUNDRY_ROOT_CHAINID
forge script script/D3MCoreDeploy.s.sol:D3MCoreDeployScript --use solc:0.8.14 --rpc-url "$ETH_RPC_URL" --sender "$ETH_FROM" --broadcast --verify
```

Fixes shellcheck issues and leaves the same format than the test file of dss-test

deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "Invalid chainid of $FOUNDRY_ROOT_CHAINID. Please set your forking environment via ETH_RPC_URL or manually by defining FOUNDRY_ROOT_CHAINID (1 or 5)."
+    exit 1;
+}
+
+[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] && echo "Deploying '$FOUNDRY_SCRIPT_CONFIG' D3M on Mainnet"
+[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Deploying '$FOUNDRY_SCRIPT_CONFIG' D3M on Goerli"
+
+rm -f out/contract-exports.env
+export FOUNDRY_ROOT_CHAINID
+forge script script/D3MDeploy.s.sol:D3MDeployScript --use solc:0.8.14 --rpc-url $ETH_RPC_URL --sender $ETH_FROM --broadcast --verify
```
gbalabasquer, line 41:
```suggestion
forge script script/D3MDeploy.s.sol:D3MDeployScript --use solc:0.8.14 --rpc-url "$ETH_RPC_URL" --sender "$ETH_FROM" --broadcast --verify
```

deploy-core.sh
gbalabasquer, line 21:
```suggestion
forge script script/D3MCoreDeploy.s.sol:D3MCoreDeployScript --use solc:0.8.14 --rpc-url "$ETH_RPC_URL" --sender "$ETH_FROM" --broadcast --verify
```

deploy-core.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -e
+
+[[ -n "$FOUNDRY_ROOT_CHAINID" ]] || {
+    [[ -n $ETH_RPC_URL ]] || {
+        echo "Please set FOUNDRY_ROOT_CHAINID (1 or 5) or ETH_RPC_URL";
+        exit 1;
+    }
+    export FOUNDRY_ROOT_CHAINID="$(cast chain-id)"
```
gbalabasquer, line 9:
```suggestion
    FOUNDRY_ROOT_CHAINID="$(cast chain-id)"
```

deploy.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -e
+
+for ARGUMENT in "$@"
+do
+    KEY=$(echo $ARGUMENT | cut -f1 -d=)
+    VALUE=$(echo $ARGUMENT | cut -f2 -d=)
```
gbalabasquer, line 7:
```suggestion
    KEY=$(echo "$ARGUMENT" | cut -f1 -d=)
    VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
```

script/D3MCoreDeploy.s.sol
hexonaut: Saves stack too big errors. Using storage doesn't really matter.

script/D3MDeploy.s.sol
```diff
@@ -0,0 +1,82 @@
...
+    DssInstance dss;
+
+    string d3mType;
+    address admin;
+    address hub;
+    bytes32 ilk;
+    D3MInstance d3m;
+
+    function run() external {
+        config = ScriptTools.loadConfig("aave");
```
gbalabasquer, line 44: Do we want to fallback here or should we actually error if there isn't a config fle env variable?

hexonaut: Yeah this is a good point. I also have an issue with automatically defaulting to mainnet here which actually messed me up when I deployed the Hub to Goerli yesterday. I think I like having these defaults for integration tests, but not for deploying.

Maybe another function on `ScriptTools` to indicate that we everything explicitly defined or error? What do you think?

gbalabasquer: You mean to have an alternative `loadConfig`?

hexonaut: Perhaps. I also want to have it revert if the CHAINID is not specified instead of just assuming mainnet.

gbalabasquer: I think we should follow one or the approach but not really having both options, as it will end up being confusing. I'm fine either way but if you already got an experience of getting messed up due to the fallback of the chain id, then might make sense to enforce always to define it.

hexonaut: Ok. How about an overload on `loadConfig()` with empty params which will revert if nothing is defined and we axe the default of chainid = 1 and always require it specified even for integration tests.

gbalabasquer: Ok perfect, sounds good. I've approved that PR.

hexonaut: updated

hexonaut:
Wanting to add another change to this: https://github.com/makerdao/dss-test/pull/25

hexonaut:
I've refactored the way export/import works. I think this is a lot cleaner.

hexonaut:
I feel like this is most of the way there now, but would still like to think if this is the best way to handle export naming, importing, etc. I'm overriding with an environment variable, but maybe it's better to just set the output name in the config json?

src/deploy/D3MDeploy.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    function deployCore(
+        address deployer,
+        address owner,
+        address daiJoin
+    ) internal returns (D3MCoreInstance memory d3mCore) {
+        d3mCore.hub = address(new D3MHub(daiJoin));
+        d3mCore.mom = address(new D3MMom());
+
+        ScriptTools.switchOwner(d3mCore.hub, deployer, owner);
```
rockyfour, line 44: this file is not a script, are you sure we should be using script tools?

script/D3MInit.s.sol
```diff
@@ -0,0 +1,123 @@
...
+    D3MCompoundConfig,
+    AavePoolLike,
+    AavePlanLike,
+    CompoundPoolLike,
+    CompoundPlanLike
+} from "../src/deploy/D3MInit.sol";
+
+contract D3MInitScript is Script {
+
+    string constant NAME = "d3m";
```
rockyfour, line 38: unused?

src/deploy/D3MDeploy.sol
hexonaut: Maybe a better name? It's just a bunch of helper functions for common tasks.

script/D3MInit.s.sol
hexonaut: fixed

src/deploy/D3MDeploy.sol
rockyfour: Can't think of anything better atm so I guess leaving it like this is also fine.

init-core.sh
```diff
@@ -0,0 +1,32 @@
...
+[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Initializing D3M Core on Goerli"
+
+export FOUNDRY_ROOT_CHAINID
+unset ETH_FROM
+cast rpc anvil_setBalance $MCD_PAUSE_PROXY 0x10000000000000000 > /dev/null
+cast rpc anvil_impersonateAccount $MCD_PAUSE_PROXY > /dev/null
+
+forge script script/D3MCoreInit.s.sol:D3MCoreInitScript --use solc:0.8.14 --rpc-url $ETH_RPC_URL --broadcast --unlocked --sender $MCD_PAUSE_PROXY
+
+cast rpc anvil_stopImpersonatingAccount $MCD_PAUSE_PROXY > /dev/null
```
gbalabasquer, line 32:
```suggestion
cast rpc anvil_setBalance "$MCD_PAUSE_PROXY" 0x10000000000000000 > /dev/null
cast rpc anvil_impersonateAccount "$MCD_PAUSE_PROXY" > /dev/null

forge script script/D3MCoreInit.s.sol:D3MCoreInitScript --use solc:0.8.14 --rpc-url "$ETH_RPC_URL" --broadcast --unlocked --sender "$MCD_PAUSE_PROXY"

cast rpc anvil_stopImpersonatingAccount "$MCD_PAUSE_PROXY" > /dev/null
```

init.sh
```diff
@@ -0,0 +1,52 @@
...
+[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Initializing '$FOUNDRY_SCRIPT_CONFIG' D3M on Goerli"
+
+export FOUNDRY_ROOT_CHAINID
+unset ETH_FROM
+cast rpc anvil_setBalance $MCD_PAUSE_PROXY 0x10000000000000000 > /dev/null
+cast rpc anvil_impersonateAccount $MCD_PAUSE_PROXY > /dev/null
+
+forge script script/D3MInit.s.sol:D3MInitScript --use solc:0.8.14 --rpc-url $ETH_RPC_URL --broadcast --unlocked --sender $MCD_PAUSE_PROXY
+
+cast rpc anvil_stopImpersonatingAccount $MCD_PAUSE_PROXY > /dev/null
```
gbalabasquer, line 52:
```suggestion
cast rpc anvil_setBalance "$MCD_PAUSE_PROXY" 0x10000000000000000 > /dev/null
cast rpc anvil_impersonateAccount "$MCD_PAUSE_PROXY" > /dev/null

forge script script/D3MInit.s.sol:D3MInitScript --use solc:0.8.14 --rpc-url "$ETH_RPC_URL" --broadcast --unlocked --sender "$MCD_PAUSE_PROXY"

cast rpc anvil_stopImpersonatingAccount "$MCD_PAUSE_PROXY" > /dev/null
```

init.sh
```diff
@@ -0,0 +1,52 @@
...
+    fi
+
+    case "$KEY" in
+            config)         export FOUNDRY_SCRIPT_CONFIG="$VALUE" ;;
+            *)
+    esac
+done
+
+[[ -n "$FOUNDRY_SCRIPT_CONFIG" ]] || {
+    echo "Please specify the D3M configration JSON. Example: ./deploy.sh config=aave";
```
gbalabasquer, line 21:
```suggestion
    echo "Please specify the D3M configration JSON. Example: ./init.sh config=aave";
```

hexonaut:
Apologies for the churn. I'm iterating on the import/export structure. This is not ideal, but I think it will be okay until Foundry provides us with higher-level dependency tooling. Simple dependencies will be common and can just be specified in the script such as d3m "core" being a dependency of deploying a d3m. More advanced stuff will need to be supplied via environment variable. You can see an example of this in `init.sh` which requires both core and d3m contracts. Would much rather this stuff be done via CLI args, but we cannot do that at the moment. Open to suggestions of alternatives.

script/D3MCoreInit.s.sol
```diff
@@ -34,19 +34,19 @@ contract D3MCoreInitScript is Script {
...
-    string deployedContracts;
+    string dependencies;
     DssInstance dss;
 
     D3MCoreInstance d3mCore;
 
     function run() external {
         config = ScriptTools.loadConfig(NAME);
-        deployedContracts = ScriptTools.readOutput(NAME);
+        dependencies = ScriptTools.loadDependencies(NAME);
```
gbalabasquer, line 44: Not sure I follow your explanation why is better use loadDependencies and the read of the file in the bash than here.

hexonaut: The problem comes in when you are integrating a d3m deploy with a higher-level deploy. If you are just working inside dss-direct-deposit the previous way was fine, but for more complex setups you need to be able to pipe through arbitrary configs that may have dependencies further up the chain.

init.sh
```diff
@@ -0,0 +1,54 @@
...
+[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] || [[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] || {
+    echo "Invalid chainid of $FOUNDRY_ROOT_CHAINID. Please set your forking environment via ETH_RPC_URL or manually by defining FOUNDRY_ROOT_CHAINID (1 or 5)."
+    exit 1;
+}
+
+[[ "$FOUNDRY_ROOT_CHAINID" == "1" ]] && echo "Initializing '$FOUNDRY_SCRIPT_CONFIG' D3M on Mainnet"
+[[ "$FOUNDRY_ROOT_CHAINID" == "5" ]] && echo "Initializing '$FOUNDRY_SCRIPT_CONFIG' D3M on Goerli"
+
+export FOUNDRY_ROOT_CHAINID
+export FOUNDRY_SCRIPT_DEPS_TEXT=`jq -sc ".[0] * .[1]" script/output/$FOUNDRY_ROOT_CHAINID/core-latest.json script/output/$FOUNDRY_ROOT_CHAINID/$D3M-latest.json`
```
gbalabasquer, line 47:
```suggestion
FOUNDRY_SCRIPT_DEPS_TEXT=$(jq -sc ".[0] * .[1]" script/output/"$FOUNDRY_ROOT_CHAINID"/core-latest.json script/output/"$FOUNDRY_ROOT_CHAINID"/"$D3M"-latest.json)
export FOUNDRY_SCRIPT_DEPS_TEXT
```

gbalabasquer (approved):
LGTM

hexonaut:
If @rockyfour is okay I will merge. The use of exports/imports will be replaced by a higher-level tool described here: https://github.com/foundry-rs/foundry/issues/3911#issuecomment-1406588724

Someone from the Foundry team is working on this, so would prefer to use official tooling rather than spend time rolling our own. The dependency management and stuff is in a working, but hacky state rn. I think it's okay to merge like this, and I'll open a new PR to update to the new `forge deploy` format when it is ready.

rockyfour (approved):
LGTM.
Didn't run the scripts but I assume you did.
Will play with it next week, I think further refinements can be done separately.

Some things we discussed:
* A future development is to use the functions in src/deploy also for the integration tests in this repo.
* The ability to replace just a single module (plan/pool) might also be needed (but might be good to gain some ground with the current scripts before we complicate them more). 
* `dss-test` content is still in flux and contains more than just testing  - so for now we will leave its usage to only a few repos.

PR #94, opened 2023-01-16
Support multiple cDai rate models and implementations

rockyfour:
This change adds the ability for the 3rd party (initially Compound) to upgrade their cDai rate model or cDai implementation without Maker having to unwind the d3m position. For that Maker governance would need to `addTack` and/or `addDelegate` with the new supported contract address, before the 3rd party change is made.

This change was not done in the past for the sake of simplicity / code convergence, and also since there were concerns for supporting the 3rd party to return to an old erroneous /malicious implementation. However, I think the concern is limited since in questionable scenarios Maker governance can always choose not to use this new ability and switch addresses the old way (while suffering some d3m downtime).

If this change is supported, an Aaave equivalent should be considered / implemented.

### DIFF
--- a/src/plans/D3MCompoundPlan.sol
+++ b/src/plans/D3MCompoundPlan.sol
@@ -79,10 +79,7 @@ contract D3MCompoundPlan is ID3MPlan {
     event Rely(address indexed usr);
     event Deny(address indexed usr);
     event File(bytes32 indexed what, uint256 data);
-    event AddTack(address indexed tack);
-    event RemoveTack(address indexed tack);
-    event AddDelegate(address indexed delegate);
-    event RemoveDelegate(address indexed delegate);
+    event File(bytes32 indexed what, address addr, uint256 data);
 
     constructor(address cDai_) {
         cDai = CErc20Like(cDai_);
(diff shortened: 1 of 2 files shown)


hexonaut (approved):
LGTM.

src/plans/D3MCompoundPlan.sol
```diff
@@ -128,14 +131,31 @@ contract D3MCompoundPlan is ID3MPlan {
...
+    }
+
+    function addDelegate(address delegate) external auth {
+        delegates[delegate] = 1;
+        emit AddDelegate(delegate);
+    }
+
+    function removeDelegate(address delegate) external auth {
+        delegates[delegate] = 0;
+        emit RemoveDelegate(delegate);
```
iamchrissmith, line 152: I'm not sure about breaking away from our `file` pattern, what about something like:

```solidity
function file(bytes32 what, address data, uint256 set) external auth {
  require(set == 0 || set == 1, "D3MCompoundPlan/file-invalid-set");
  if (what == "tack") tacks[data] = set;
  else if (what == "delegate") delegates[data] = set;
  else revert(""D3MCompoundPlan/file-file-unrecognized-param");
  emit File(what, data, set);
}
```
or 
```solidity
    function file(bytes32 what, address data) external {
        uint256 set;
        if (what == "tack") {
            set = ~tacks[data] & 1;
            tacks[data] = set;
        } else if (what == "delegate") {
            set = ~delegates[data] & 1; 
            delegates[data] = set;
        } else revert("D3MCompoundPlan/file-unrecognized-param");
        emit File(what, data, set);
    }
```
or
```solidity
    function file(bytes32 what, address data) external {
        uint256 set;
        if (what == "tack") {
            set = (tacks[data] == 1) ? 0 : 1;
            tacks[data] = set;
        } else if (what == "delegate") {
            set = (delegates[data] == 1) ? 0 : 1;
            delegates[data] = set;
        } else revert("D3MCompoundPlan/file-unrecognized-param");
        emit File(what, data, set);
    }
```

rockyfour: Yeah, I like your 1st option, though haven't seen we did that with a boolean-like value before. Let's see what others think.

iamchrissmith: > though haven't seen we did that with a boolean-like value before.

@talbaneth i'm not sure what you mean by this?

and I guess in addition to the example Sam has here, the best argument against what I'm proposing here is our `rely`/`deny` structure...

If others are okay with this I'm fine leaving as-is

gbalabasquer: I'd go for option 1 but changing `data` to `addr` and `set` to `data`. This is probably the most similar example we have: https://github.com/makerdao/dss-teleport/blob/master/src/TeleportRouter.sol#L85

iamchrissmith: > I'd go for option 1 but changing `data` to `addr` and `set` to `data`. This is probably the most similar example we have: https://github.com/makerdao/dss-teleport/blob/master/src/TeleportRouter.sol#L85

yeah `addr` and `data` is better

rockyfour: Done in [6754ed0](https://github.com/makerdao/dss-direct-deposit/pull/94/commits/6754ed0a1bf5a0fb481c75a5f4ce3dafd0c024cf)

hexonaut (approved):
New changes look fine to me.

iamchrissmith (approved):
changes lgtm

PR #97, opened 2023-02-24
Actual pools and plans renaming

gbalabasquer:


hexonaut (commented):
Changes look good. If we are generalizing to the code away from specific instances I would remove all references to smart contract deployments inside the code such as https://github.com/makerdao/dss-direct-deposit/blob/master/src/plans/D3MCompoundPlan.sol#L49

Those should be replaced with code references or dropped IMO.

After that this is good to merge.

hexonaut (approved):
LGTM.

iamchrissmith (approved):
I'm still not 100% sold on this balance of abstraction and specificity, but since it seemed to be group consensus, these changes lgtm

rockyfour (approved):
LGTM

src/plans/D3MCompoundV2TypeRateTargetPlan.sol
```diff
@@ -46,7 +46,7 @@ pragma solidity ^0.8.14;
 
 import "./ID3MPlan.sol";
 
-// cDai - https://etherscan.io/address/0x5d3a536E4D6DbD6114cc1Ead35777bAB948E3643
+// cDai - https://github.com/compound-finance/compound-protocol/blob/master/contracts/CErc20.sol
```
The-Arbiter, line 49: Personally I would use a commit hash rather than master here.

Raw link:
https://raw.githubusercontent.com/compound-finance/compound-protocol/a3214f67b73310d547e00fc578e8355911c9d376/contracts/CErc20.sol

Github link:
https://github.com/compound-finance/compound-protocol/blob/a3214f67b73310d547e00fc578e8355911c9d376/contracts/CErc20.sol

The-Arbiter: If the concern is that this cDai address may be deprecated in the future then we can stick to the method you've used here, but it isn't consistent with how we link to github stuff for things like execs in spells. Just my 2c

src/plans/D3MCompoundV2TypeRateTargetPlan.sol
```diff
@@ -55,7 +55,7 @@ interface CErc20Like {
     function implementation() external view returns (address);
 }
 
-// JumpRateModelV2 - https://etherscan.io/address/0xfb564da37b41b2f6b6edcc3e56fbf523bd9f2012
+// JumpRateModelV2 - https://github.com/compound-finance/compound-protocol/blob/master/contracts/BaseJumpRateModelV2.sol
```
The-Arbiter, line 58: Same thing for all GH links

The-Arbiter:
'Type' and D3M plan and pool changes look OK
File name changes OK
Formatting changes OK

_Re: [Replace specific deployments with repos references](https://github.com/makerdao/dss-direct-deposit/pull/97/commits/44dd2e38f313640ef4e8852a7f14cef130215968) I question why we are not using commit hashes here (we are moving from directing the user to an ES contract code section to a GH code URL, but we are just pointing them to master). 
Up to you guys on this one but I would anchor it to the commit that matches the deployed code._

The-Arbiter (approved):
As mentioned in the other comment, the 'master' in our URLs points to potentially dynamic code when we were previously referring to (more) static code (EtherScan contract source).

My actions here would be to ensure that the code we are linking to matches the deployed code we were linking to before, and then using a URL with the commit hash as we do with spell execs.

Let me know if this sits well with you. I'm happy to do the code diff work to ensure it all matches.

src/plans/D3MCompoundV2TypeRateTargetPlan.sol
gbalabasquer: Yeah I was thinking on doing the commit hash as well, but didn't have the time to go over one by one to make extremely sure it was all perfectly matching. So left it as a soft reference.

We have a few options:
1) Leave as it is
2) Put a comment making explicit is a soft reference
3) Go contract by contract and check the implementation is exactly that and then hardening the commit (if you want to do that sounds great to me).

hexonaut: I thought of this as well, and personally think it's fine to leave as a soft reference in this instance.

gbalabasquer: We can create a new PR if this change wants to be pursue, opting for option 1 for now.

The-Arbiter: This is fine then as long as we know what we're doing

PR #98, opened 2023-02-24
Refactor scripts to split out pools and plans for more flexilibility

hexonaut:
A precursor to having pools and plans mix and match. For example you may want a different targeting mechanism for an Aave pool adapter.

I've also fixed the json config variables to use the `.` at the beginning syntax and using Compound Plan's new multiple `tack`/`delegate` setup.

Please note the code is for V2 of Aave, but the initialization logic is agnostic between V2 and V3. I'm not sure about Compound V2 vs V3 though.

Can wait to merge this after the pool/plan renaming is finished.


script/D3MCoreDeploy.s.sol
```diff
@@ -41,7 +41,7 @@ contract D3MCoreDeployScript is Script {
 
     function run() external {
         config = ScriptTools.loadConfig(NAME);
-        dss = MCD.loadFromChainlog(config.readAddress("chainlog"));
+        dss = MCD.loadFromChainlog(config.readAddress(".chainlog"));
 
         admin = config.readAddress("admin");
```
The-Arbiter, line 46: Looks like this one got missed when '.' was prepended to the config field names

script/D3MDeploy.s.sol
```diff
@@ -57,35 +58,66 @@ contract D3MDeployScript is Script {
...
                 admin,
                 ilk,
-                address(dss.vat),
                 hub,
-                config.readAddress("cdai")
+                config.readAddress(".cdai")
             );
         } else {
-            revert("unknown-d3m-type");
+            revert("Unknown pool type");
```
The-Arbiter, line 98: Should this be:
```suggestion
            revert("Unknown-pool-type");
```
With dashes?

script/D3MDeploy.s.sol
```diff
@@ -57,35 +58,66 @@ contract D3MDeployScript is Script {
...
+                    config.readAddress(".lendingPool")
+                );
+            } else if (poolType.eq("compound-v2")) {
+                d3m.plan = D3MDeploy.deployCompoundV2TypeRateTargetPlan(
+                    msg.sender,
+                    admin,
+                    config.readAddress(".cdai")
+                );
+            } else {
+                revert("Invalid pool type for rate target plan type");
```
The-Arbiter, line 117: Same thing for the error here

```suggestion
                revert("Invalid-pool-type-for-rate-target-plan-type");
```

script/D3MDeploy.s.sol
```diff
@@ -57,35 +58,66 @@ contract D3MDeployScript is Script {
...
+            }
+        } else {
+            revert("Unknown plan type");
```
The-Arbiter, line 120:
```suggestion
            revert("Unknown-plan-type");
```

script/D3MDeploy.s.sol
The-Arbiter: This is how it was before. If we are moving away there should be a good reason IMO

src/deploy/D3MInit.sol
```diff
@@ -187,16 +200,17 @@ library D3MInit {
         );
         dss.spotter.poke(ilk);
 
+        GemAbstract gem = GemAbstract(ID3MPool(d3m.pool).redeemable());
```
The-Arbiter, line 203: Require this is non zero? From PR#93 (same discussion) but we could require gem.decimals() and symbol are also not 0? 

Up to you

The-Arbiter (approved):
Apologies if this is a premature review

Changes to '.chainlog' and other prepending of '.' OK (one was missed).
*Can I ask why we need to do this now if the config.readAddress() function hasn't changed and the template json's haven't either. Maybe I am missing something here.*

Error / revert stuff without dashes: Up to consensus but I would stick with our existing patterns unless there's a reason not to. Up to you guys.

Other renaming and split of config pulls from D3MInit OK
WAD add OK

D3MDeploy changes (split of deploys by pool plan and oracle) OK
Change from single value to mapping for tacks() and delegates() OK and tests for this OK

D3M Plan relying on Mom from config OK

Shift from cDai values in code to config file OK (script code looks OK)

---

My primary question here (for myself) is why we need to change 'chainlog' to '.chainlog'.

For things like cDai config, I am assuming that this is yet to be added to the code here?

script/D3MCoreDeploy.s.sol
hexonaut: Good catch. Fixed.

script/D3MDeploy.s.sol
hexonaut: Previously it AI-autocompleted to that dash format, but I later changed my mind. No need for it to be cryptic as this will just be the error the deployer sees. It can be in plain English.

src/deploy/D3MInit.sol
hexonaut: Personally I think it's okay as if the address is 0 or really any non-erc20-like contract it will revert when attempting to get the decimals().

hexonaut:
> Can I ask why we need to do this now if the config.readAddress() function hasn't changed and the template json's haven't either. Maybe I am missing something here.

This was a fix made to forge-std where we the non-dot prefixed version was actually a bug that they've fixed. Newer versions of foundry will error if you don't prepend the dot (which means root json object). This is `jq` syntax.

> For things like cDai config, I am assuming that this is yet to be added to the code here?

What do you mean by this? I'm not clear.

rockyfour (approved):
LGTM

gbalabasquer (approved):
LGTM

PR #95, opened 2023-02-05
Add Debt Ceiling Plan

hexonaut:
Simplest plan which just targets the debt ceiling. This will be useful for MIP65-like constructs.

PR #99, opened 2023-03-01
Refactor scripts

iamchrissmith:
Separating scripts (deploy and init) so that as there are more D3Ms we can better manage these without the conditionals.


iamchrissmith:
@gbalabasquer and @hexonaut this is was I was referring to with regards to the deploy/init scripts.  As there are future D3Ms added, I think having separate files/contracts for the deploy and init in addition to the config will be helpful and make reviews easier.

hexonaut:
So I get the instinct here, and it will work with this design under the current setup which combines pool and plan deploys in one action, but this is being changed to split out the pool and plans in this PR: https://github.com/makerdao/dss-direct-deposit/pull/98

If you are deploying an instance to connect to an instance of the Aave codebase then you may not want to use the rate targeting plan and instead the fixed buffer plan or perhaps one that just targets the vat debt ceiling here: https://github.com/makerdao/dss-direct-deposit/pull/95/files#diff-2116aabfda758923c25291f0bcebe2d64df2cf7ba93f5e35ef4d6bbeff34ecd1 . As such you need to have the deploy script able to flexibly define which plan is going with the pool adapter and this doesn't cater well to splitting the files out because plans like the vat debt ceiling one can be used by any D3M.

If you really feel strongly to split the files I suppose the common plan deploy logic could be moved to a shared contract or something, but that seems a bit messy to me. Otherwise it will be a lot of repetitive code.

iamchrissmith:
So I think deploy scripts should be specific to the deploys more than a single menu for how to deploy any conceivable D3M. 

So our current deploy situation would be:

DeployAaveV2 uses:
- D3MAaveV2TypeRateTargetPlan
- D3MAaveV2TypePool

DeployCompoundV2 uses:
- D3MCompoundV2TypeRateTargetPlan
- D3MCompoundV2TypePool

Eventually your SparkLend D3M would uses something like: 
- D3MBufferTargetPlan
- D3MAaveV3TypePool

This will allow us to know how each of the D3Ms were actually deployed, maybe we need to change the bash scripts for this and tie them to the `ilk`s we're setting up, but that was my thinking with this PR.

> If you really feel strongly to split the files I suppose the common plan deploy logic could be moved to a shared contract or something, but that seems a bit messy to me. Otherwise it will be a lot of repetitive code.

I'm not sure I follow what you're saying here: Shared contract is messy? Are you thinking like a library or something that gets deployed and the other deploys use?

iamchrissmith:
Yeah I see what you're saying.  I think I'm looking at it from the reviewer perspective and trying to minimize the number of overlapping changes we have and different contexts one needs to load to understand what is happening. 

Since deploys are kind of one time/specific things. My thinking is that the deploys are more like database migrations.  Plus that opens up a better way (possibly) of requiring, managing and organizing the off boarding script stuff that Mooney has asked about.  If we go that route maybe we should have the organization like:

|- src
|    - deploys
|        - D3MDeployBase.s.sol
|        - 20230306-AaveV2 (maybe named after the ilk?)
|            - Deploy.s.sol
|            - Init.s.sol
|            - Offboard.s.sol
|            - aavev2.json

an outside contributor could then (assuming they are reusing existing plans/pools) could just submit a new deploy directory and it would contain everything core reviewers needed to understand what was being deployed and how in one place.

If we want to go this way, this PR probably needs to be updated and go further in re-organizing things.

gbalabasquer:
Hmmm I don't think we should use project repos for saving specific configurations. We had this same sort of discussion some time ago with the input files, which in my opinion (and I believe @rockyfour as well) should be just templates, specific deployment configuration might go into a different repo (if we actually want to save them in a repo).
This version is going a step above and actually forcing the script per se to set specific configurations, so I'd feel is not the best path to follow.
So I think I prefer Sam's PR. However I would go one step further trying to avoid the conditionals at all (as they also do not feel good), we might be able to do all the conditional thing in the input file itself, so something like defining there which `deployPool/Plan` function we will choose and let the script solidity file to build and execute from there. If we use plain `.call` we should be able to build the calldata dynamically from what we bring from the input file.

On the offboarding case, I think we should think similarly, a function to offboard a specific pool/plan type, and then having the spell to call the one that we need.

I might be missing details (maybe important ones), but that's how I see it at high level.

iamchrissmith:
ok

PR #102, opened 2023-03-08
Add Support for Spark Lend

hexonaut:
Adds the Pool and Plan as specified by https://forum.makerdao.com/t/mip116-d3m-to-spark-lend/19732 with an integration test against the mainnet deploy of Spark Lend.

Please note the following:

 * The buffer plan is generic across both V2 and V3 of Aave.
 * Reserve data format changed between V2 and V3 for the pool adapter
 * Rewards collection interface changed between V2 and V3 for the pool adapter (Spark does not use rewards, but made sense to update the adapter in case this changes in the future)
 * V3 introduces supply caps that need to be accounted for in the V3 pool - the adapter uses a no-op flash loan to make sure the `liquidityIndex` and `accruedToTreasury` values are not stale. This is a bit hacky, but was the best solution available to trigger an index update. I opted to write changes to the indices because reading in all the values (instead of triggering an update) would be incredibly brittle. (Spark does not use supply caps so this part is ignored, but will be useful if they are ever turned on or the pool is used in another instance where supply cap is turned on)
 * Integration test is fairly simple compared to others because the general D3M functionality is already tested. Maybe as an improvement the general D3M integration test logic could be pulled into a base contract to ensure correctness on every D3M implementation.


hexonaut:
This is ready for review now.

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    function preDebtChange() external override {
+        // Execute a 1 wei flash loan to trigger an index update on everything
+        // If the fee is less than 100% than the fee should always round down to zero
+        // Note: we can ignore indicies being up to date if the supply cap is 0
+        PoolLike.ReserveData memory data = pool.getReserveData(address(dai));
+        if (
+            // The next two lines can be slightly optimized, but it's easier to read like this
+            (data.configuration & ~FLASHLOAN_ENABLED_MASK) != 0 &&
```
rockyfour, line 233: What would be the implications if flash loans are disabled?

rockyfour: Also why do we need to update the indexes here vs the v2 d3m?

hexonaut: I added success and failure tests to cover this case.

src/plans/D3MAaveTypeBufferPlan.sol
```diff
@@ -0,0 +1,111 @@
...
+            buffer = data;
+        } else revert("D3MAaveTypeBufferPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    function getTargetAssets(uint256 currentAssets) external override view returns (uint256) {
+        if (buffer == 0) return 0; // De-activated
+
+        uint256 liquidityAvailable = dai.balanceOf(address(adai));
+        if (buffer >= liquidityAvailable) {
```
iamchrissmith, line 84: I find this naming confusing.  To me the buffer is the additional amount, but the way it is used here is the target total liquidity.  This and other parts of this function could be reworked to say `if (currentAssets < liquidityAvailable + buffer) { // add liquidity`.  I think that would make the code more understandable and provide a better way for governance/risk to set and understand the `buffer` setting

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+            vat.nope(hub);
+            hub = data;
+            vat.hope(data);
+        } else if (what == "king") king = data;
+        else revert("D3MAaveV3TypePool/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    // Deposits Dai to Aave in exchange for adai which is received by this contract
+    // Aave: https://docs.aave.com/developers/v/2.0/the-core-protocol/lendingpool#deposit
```
iamchrissmith, line 192: this should be pointed to the v3 docs

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+        pool.deposit(address(dai), wad, address(this), 0);
+
+        // Verify the correct amount of adai shows up
+        uint256 interestIndex = pool.getReserveNormalizedIncome(address(dai));
+        uint256 scaledAmount = _rdiv(wad, interestIndex);
+        require(adai.scaledBalanceOf(address(this)) >= (scaledPrev + scaledAmount), "D3MAaveV3TypePool/incorrect-adai-balance-received");
+    }
+
+    // Withdraws Dai from Aave in exchange for adai
+    // Aave: https://docs.aave.com/developers/v/2.0/the-core-protocol/lendingpool#withdraw
```
iamchrissmith, line 205: this should point to the v3 docs

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    function quit(address dst) external override auth {
+        require(vat.live() == 1, "D3MAaveV3TypePool/no-quit-during-shutdown");
+        require(adai.transfer(dst, adai.balanceOf(address(this))), "D3MAaveV3TypePool/transfer-failed");
+    }
+
+    function preDebtChange() external override {
+        // Execute a 1 wei flash loan to trigger an index update on everything
+        // If the fee is less than 100% than the fee should always round down to zero
```
iamchrissmith, line 228: typo:
```suggestion
        // If the fee is less than 100% then the fee should always round down to zero
```

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+        if (
+            // The next two lines can be slightly optimized, but it's easier to read like this
+            (data.configuration & ~FLASHLOAN_ENABLED_MASK) != 0 &&
+            (data.configuration & ~SUPPLY_CAP_MASK) != 0 &&
+            dai.balanceOf(address(adai)) > 0
+        ) {
+            pool.flashLoanSimple(address(this), address(dai), 1, "", 0);
+        }
+    }
+    function executeOperation(
```
iamchrissmith, line 240: add line

src/pools/D3MAaveV3TypePool.sol
```diff
@@ -0,0 +1,292 @@
...
+    event File(bytes32 indexed what, address data);
+    event Collect(address indexed king, address indexed gift, uint256 amt);
+
+    constructor(bytes32 ilk_, address hub_, address dai_, address pool_) {
+        ilk = ilk_;
+        dai = TokenLike(dai_);
+        pool = PoolLike(pool_);
+
+        // Fetch the reserve data from Aave
+        PoolLike.ReserveData memory data = pool.getReserveData(dai_);
```
iamchrissmith, line 127: in the aave pool and places were we access ilk/urn data we grab the specific variables we want rather than using a generic memory variable.  I think that is more explicit and cleaner as it avoids a bunch of `data.` references later

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,335 @@
...
+    }
+    function _divup(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        unchecked {
+            z = x != 0 ? ((x - 1) / y) + 1 : 0;
+        }
+    }
+    function getSupplyUsed() internal view returns (uint256) {
+        PoolLike.ReserveData memory data = sparkPool.getReserveData(address(dss.dai));
+        return _divup((ATokenLike(adai).scaledTotalSupply() + uint256(data.accruedToTreasury)) * data.liquidityIndex, RAY);
+    }
```
iamchrissmith, line 235: needs spacing between functions or they blend together

src/pools/D3MAaveV3TypePool.sol
rockyfour: I think I like your suggestion to drop the supply cap support, which would narrow the problem space.
We can have a supply cap check in active() and unwind if there is one.
It can be tackled again when AaveV3 pools with supply caps are supported.

I'm not 100% sure but I don't think it would require changing to a new name (e.g `SparkType`), since once supply caps are supported it will just be considered an upgrade and will work also for Spark.

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: The proposed change would be different logic though. `liquidityAvailable` is the amount of DAI in the pool and that is what this Plan attempts to make a fixed value. So say if `buffer` is `50m` then it will try to keep the amount of dai balance in the pool at `50m`.

I can rename the variables if it's confusing, but `liquidityAvailable` is pretty clear to me it's the amount of liquid DAI (IE not being borrowed). `Total market size = Total Debt + Liquidity Available`

LMK if you have better alternatives to names, but the logic should be correct as-is.

src/pools/D3MAaveV3TypePool.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/5eb5703a0abc361073e4eabe4e111cf31863e37b

src/pools/D3MAaveV3TypePool.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/5eb5703a0abc361073e4eabe4e111cf31863e37b

src/pools/D3MAaveV3TypePool.sol
hexonaut: There is stack too deep using the V2 method. I mention this in the top of the file.

src/tests/integration/SparkLend.t.sol
hexonaut: Person preference to sometimes group closely related functions, but it's okay I'll add the lines.

src/pools/D3MAaveV3TypePool.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/c84bc4be70c2f8640a6a77eebc95529bd0f5a0d5

src/pools/D3MAaveV3TypePool.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/c84bc4be70c2f8640a6a77eebc95529bd0f5a0d5

src/tests/integration/SparkLend.t.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/c84bc4be70c2f8640a6a77eebc95529bd0f5a0d5

hexonaut:
> We have not tested core D3M functionality against an AaveV3Type implementation (only V2) and we have not tested it against a buffer plan (only a rate plan). Both of these changes could have implications for debt ceiling interactions, fee collection, liquid crunches, etc so I think it is important to have them in this integration test. I'm fine if you think it is better to pull them into a base contract that gets inherited here, but I don't think the current approach of just not having them works.

Fair point. I'll build a generic base integration test contract.

src/plans/D3MAaveTypeBufferPlan.sol
iamchrissmith: ah ok, that sounds more like a targetLiq or targetAmt.  As I said, buffer is the cushion amount so imo it creates confusion calling the total a buffer.  To try to keep some of the maker memes, maybe `mark` or `goal`

src/tests/integration/SparkLend.t.sol
iamchrissmith: > Person preference to sometimes group closely related functions

In alignment with Rune's goals of having "rules" that SubDAOs and outside contributors should be following on submitted code, we created the PE core guidelines and in anything not covered in there said we should defer to Solidity style guides.  In this case, the Style Guide does explicitly say we [should have spaces](https://docs.soliditylang.org/en/v0.8.17/style-guide.html#blank-lines) and not do this.

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: But the total is not `buffer`. In the example above `buffer == 50m`, and it's seeing if the `liquidityAvailable` is above or below `50m` (the buffer). Adding the `currentAssets` in the return is just because the function return expects it in that absolute number format.

src/pools/D3MAaveV3TypePool.sol
gbalabasquer: Yes I consider the same than Chris, if AaveV3 official implementation uses the cap, and Spark not, I tend to rename it to `SparkTypePool`. Now if AaveV3 doesn't currently use caps either and will start using it in the future, maybe we could name it a `AaveV3TypePoolNoCap`.

hexonaut: I really think it will be more confusing to mix instance vs code for naming. The codebase is still Aave V3 even if for Spark Lend the caps are turned off. What if someone else deploys another instance of Aave V3 with the caps turned off? Does it make sense to use a `SparkTypePool` adapter? There's no relation there, but it's the same code.

I think as a compromise I'd be okay with adding the "NoCap" to the end so it is `AaveV3NoSupplyCapTypePool` and splitting out the cap support into a new PR (which doesn't have to be looked at until it's needed). How does that sound?

gbalabasquer: is it a code change what you are doing in Spark for removing the caps or just setting a value to max?

hexonaut: > is it a code change what you are doing in Spark for removing the caps or just setting a value to max?

There is no code changes between the official Aave V3 instance and Spark Lend. The only difference is configuration.

gbalabasquer: Ok then I think you are probably right and the correct name should be like `AaveV3NoSupplyCapTypePool`.

src/plans/D3MAaveTypeBufferPlan.sol
```diff
@@ -0,0 +1,111 @@
...
+                    return 0;
+                }
+            }
+        }
+    }
+
+    function active() public view override returns (bool) {
+        if (buffer == 0) return false;
+        uint256 adaiRevision_ = adai.ATOKEN_REVISION();
+        return adaiRevision_ == adaiRevision;
```
gbalabasquer, line 103:
```suggestion
        return buffer > 0 && adai.ATOKEN_REVISION() == adaiRevision;
```

gbalabasquer: Btw did we consider removing the adai revision and make this plan more standard?
I guess that was your first idea as you had mentioned it wouldn't be specific for any project, but then something changed that this was considered necessary.

hexonaut: Yeah I'm not sure if it adds much. Lots of things can change and really all it needs to do is look like a rebasing ERC-20. Any change that would break the assumptions here would be the same as if Aave put through a malicious change or something (IE breaking everything).

gbalabasquer: Yeah for the purpose of the plan, we just need a "source" address where to check the actual liquidity available we don't care what that token is. And if we want to check some malicious thing, then that adai_revision check is quite poor. So not sure, I would tend to remove it and do something standard for any project.

src/pools/D3MAaveV3TypePool.sol
iamchrissmith: Seems like you all are in agreement, but I'll just weigh in here and to say again that this feels to me like we're getting the worst of both specificity and abstraction.  we have a very specific "abstract" implementation for a certain configuration of code and i'm not really sure how its better than just calling the code after the instance its pointed to.   If we follow this logic, i'm worried we'll end up with something like `AaveV3NoSupplyCapNonRebasingStableRateDaiOnlyTypePool` in the future which just seems like the wrong place to end up. 

Plus if we go this route and actually do find two use cases for our "abstract" `AaveV3NoCapTypePool` and then some other configuration changes between those two we'll be right back to this debate of adding contextual/complex code to the pool or separating them again to a configuration abstraction (see naming above).

I would vote for 
1) leave this abstract `AaveV3TypePool` with the code so it works against the Aave code base regardless of how it is configured
or
2) build a `SparkPool` that works for the D3M we are building that governance has decided we should build and then have an AaveV3Pool or FluxPool or WhateverSubDAOPool in the future when we know what that looks like

With a preference for 2 because 1 feels like we're trying to optimize for a future we don't really know yet and paying a complexity price today for unknown benefit...

src/plans/D3MAaveTypeBufferPlan.sol
iamchrissmith: ok, yes I see what you're saying and how I misinterpreted it. So `buffer` is the target liquidity Maker Governance wants in the your spark system. I think with that understanding i'm fine keeping the name as is, but also feel like it kind of proves the value of daiwanese to avoid my confusion preconceptions... maybe `buff` is a good middle ground, but i'd say unless @gbalabasquer or @rockyfour want to tip the scale for a change, lets just leave it

gbalabasquer: `buff` seems a good name and shouldn't be that problematic to change it.

src/plans/D3MAaveTypeBufferPlan.sol
gbalabasquer: @iamchrissmith @rockyfour opinions on this one before thinking on any change?

hexonaut:
I've added a generic integration base with the full test suite of Aave V2. This should be pretty general across D3Ms.

hexonaut:
Supply cap is removed now as well. If it is desired I'll remove the adai revision stuff too.

src/pools/D3MAaveV3TypePool.sol
gbalabasquer: > i'm worried we'll end up with something like `AaveV3NoSupplyCapNonRebasingStableRateDaiOnlyTypePool` in the future which just seems like the wrong place to end up.

Yes this situation is not good either. To be honest I don't think there is a perfect solution, as we discussed in the very beginning. All of them will have pros and cons. I still feel better the way it was recently pushed because describes more the code than the project where it will used, but yeah certainly if we end up in a situation as you put as example that will be very bad as well.

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: Any issue with just leaving it? `buffer` and `buff` are the same thing to me. :)

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
```diff
@@ -0,0 +1,253 @@
...
+        z = (x * RAY) / y;
+    }
+    function _min(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = x <= y ? x : y;
+    }
+    function _divup(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        unchecked {
+            z = x != 0 ? ((x - 1) / y) + 1 : 0;
+        }
+    }
```
gbalabasquer, line 162: This on is not used without the cap limit.

src/tests/pools/D3MAaveV3NoSupplyCapTypePool.t.sol
```diff
@@ -0,0 +1,329 @@
...
+        assertEq(amt, 1);
+        assertEq(dst, hub);
+    }
+
+    function test_collect_claims_for_king() public {
+        address king = address(123);
+        address rewardsClaimer = adai.getIncentivesController();
+        D3MAaveV3NoSupplyCapTypePool(d3mTestPool).file("king", king);
+
+        D3MAaveV3NoSupplyCapTypePool(d3mTestPool).collect(address(123));
```
gbalabasquer, line 251: It feels a bit confusing the `reward` is the same address than the `king`, I know it doesn't matter for the test to pass, but semantically is confusing. I'd just use another random address.

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,420 @@
...
+    TreasuryAdminLike treasuryAdmin;
+    uint256 buffer;
+
+    D3MAaveTypeBufferPlan plan;
+    D3MAaveV3NoSupplyCapTypePool pool;
+
+    uint256 constant FLASHLOAN_ENABLED_MASK = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF7FFFFFFFFFFFFFFF;
+    uint256 constant SUPPLY_CAP_MASK = 0xFFFFFFFFFFFFFFFFFFFFFFFFFF000000000FFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
+    uint256 constant SUPPLY_CAP_START_BIT_POSITION = 116;
+    string constant SUPPLY_CAP_EXCEEDED = '51';  // SUPPLY_CAP_EXCEEDED error code
```
gbalabasquer, line 137: This should be removed, but maybe we can wait until be 100% sure that the supply cap is not happening.

hexonaut: I've already removed the code. Can reverse the commits if need be, but it seems everyone is leaning towards no supply cap on this one.

src/plans/D3MAaveTypeBufferPlan.sol
rockyfour: Yeah I guess I agree with the reasoning and it can be removed.

hexonaut: Ok will remove.

hexonaut: Done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/991d96935aa90ddf25b7f8e7e70cabb17c45701b

src/plans/D3MAaveTypeBufferPlan.sol
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ *  @title D3M Aave Buffer Plan
+ *  @notice Ensure `buffer` amount of DAI is always available to borrow.
+ *  @dev This plan can be used with both V2 and V3 versions of the Aave codebase.
+ */
+contract D3MAaveTypeBufferPlan is ID3MPlan {
+
+    mapping (address => uint256) public wards;
+    uint256                      public buffer;
```
rockyfour, line 37: Might be nice to specify the units in a comments, like `// WAD`

src/plans/D3MAaveTypeBufferPlan.sol
```diff
@@ -0,0 +1,106 @@
...
+        uint256 liquidityAvailable = dai.balanceOf(address(adai));
+        if (buffer >= liquidityAvailable) {
+            // Need to increase liquidity
+            return currentAssets + (buffer - liquidityAvailable);
+        } else {
+            // Need to decrease liquidity
+            unchecked {
+                uint256 decrease = liquidityAvailable - buffer;
+                if (currentAssets >= decrease) {
+                    return currentAssets - decrease;
```
rockyfour, line 89: Mostly informational, and I know this is relevant also for the existing d3ms, but note that it would be even easier to cause full unwind here with a flash loan (no need to calculate rates and amounts are pretty clear).

An existing lender can get very favorable rates by causing the d3m to unwind on will.
I guess it's a matter of how often keepers are willing to call `exec` to re-wind.

hexonaut: This is a good point. In Spark's case the rate curve is flat so there is there is really no reason to do this, but in general it's a fair point.

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/3297f7420813bdb744599ecd49c90e6440e748a8

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: Fixed https://github.com/makerdao/dss-direct-deposit/pull/102/commits/3297f7420813bdb744599ecd49c90e6440e748a8

rockyfour: > This is a good point. In Spark's case the rate curve is flat so there is there is really no reason to do this, but in general it's a fair point.

Ah right, I thought about the flat curve last week in this regard but forgot it today for some reason.

src/plans/D3MAaveTypeBufferPlan.sol
gbalabasquer: Now it is removed, we could actually just make this a standard plan, where we remove the `adai` reference and we pass in the constructor `dai` and a `gem` (`adai` for this implementation).
This way the plan would be just `D3MBufferPlan`

hexonaut: It is still pretty specific to Aave in that they store their measure of liquidity inside the aToken.

gbalabasquer: Why not call it `src`. So basically what you need is the source where to check liquidity. You do not care if it is a token or what. In this case will be `adai` but in other could be a pool address.

hexonaut: I'm saying calling `dai.balanceOf(src)` is still assuming the dai balance == available liquidity. It's possible there is additional accounting.

gbalabasquer: I mean, my main idea to remove that validation was to make this general. But if you feel strongly about keeping as aavetype I wouldn't fight for the opposite either.

gbalabasquer: I think you are right, wouldn't work even for Compound, as the totalPoolSize is defined as the balance of dai in cdai + borrows/debt, but subtracts the totalReserves. So that should also be affecting the available liquidity. Let's leave it as it is.

src/plans/D3MAaveTypeBufferPlan.sol
gbalabasquer: Yeah I guess if Chris doesn't want to pursue with it, I'd be fine leaving as it.

src/plans/D3MAaveTypeBufferPlan.sol
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+
+    function active() public view override returns (bool) {
+        return buffer > 0;
+    }
+
+    function disable() external override {
+        require(wards[msg.sender] == 1 || !active(), "D3MAaveTypeBufferPlan/not-authorized");
```
gbalabasquer, line 103: The `|| !active` isn't really necessary here as the only way to be true is because the buffer is already 0. We can choose to remove it to make it more sense, but also keep it just for symmetry and avoiding to miss it in a future plan that actually does something.

hexonaut: Good point. I'm okay with either.

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
```diff
@@ -0,0 +1,248 @@
...
+interface D3mHubLike {
+    function vat() external view returns (address);
+    function end() external view returns (EndLike);
+}
+
+interface EndLike {
+    function Art(bytes32) external view returns (uint256);
+}
+
+interface ATokenLike is TokenLike {
```
gbalabasquer, line 43:
```suggestion
// https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/tokenization/AToken.sol
interface ATokenLike is TokenLike {
```

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
```diff
@@ -0,0 +1,249 @@
...
+    function Art(bytes32) external view returns (uint256);
+}
+
+// https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/tokenization/AToken.sol
+interface ATokenLike is TokenLike {
+    function scaledBalanceOf(address) external view returns (uint256);
+    function getIncentivesController() external view returns (address);
+}
+
+interface PoolLike {
```
gbalabasquer, line 49:
```suggestion
// https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/pool/Pool.sol
interface PoolLike {
```

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
```diff
@@ -0,0 +1,250 @@
...
+        uint128 isolationModeTotalDebt;
+    }
+    
+    function supply(address asset, uint256 amount, address onBehalfOf, uint16 referralCode) external;
+    function withdraw(address asset, uint256 amount, address to) external;
+    function getReserveNormalizedIncome(address asset) external view returns (uint256);
+    function getReserveData(address asset) external view returns (ReserveData memory);
+}
+
+interface RewardsClaimerLike {
```
gbalabasquer, line 92:
```suggestion
// https://github.com/aave/aave-v3-periphery/blob/master/contracts/rewards/RewardsController.sol
interface RewardsClaimerLike {
```

src/plans/D3MAaveTypeBufferPlan.sol
gbalabasquer: I think I'd slightly lean to go for removing it. @rockyfour any opinion?

rockyfour: I'd leave it because of the reasons mentioned.
If `active()` gets updated there's a good chance re-including it in `disable()` will get forgotten.

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
```diff
@@ -0,0 +1,251 @@
...
+
+pragma solidity ^0.8.14;
+
+import "./ID3MPool.sol";
+
+interface TokenLike {
+    function balanceOf(address) external view returns (uint256);
+    function approve(address, uint256) external returns (bool);
+    function transfer(address, uint256) external returns (bool);
+    function decimals() external view returns (uint8);
```
rockyfour, line 25:
```suggestion
```
This can be removed now (it doesn't appear in the other pools).

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -0,0 +1,158 @@
...
+
+    function test_file() public {
+        checkFileUint(d3mTestPlan, contractName, ["buffer"]);
+    }
+
+    function test_auth_modifier() public {
+        plan.file("buffer", 1);
+        plan.deny(address(this));
+
+        checkModifier(d3mTestPlan, "D3MAaveTypeBufferPlan/not-authorized", [
```
rockyfour, line 83: Ah first time I see this `checkModifier`.
Is it better than using `vm.expectRevert`?

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -0,0 +1,158 @@
...
+
+    function test_increase_liquidity() public {
+        plan.file("buffer", 100 ether);
+        assertEq(plan.getTargetAssets(20 ether), 120 ether);
+    }
+
+    function test_decrease_liquidity_sole_provider() public {
+        plan.file("buffer", 100 ether);
+        assertEq(plan.getTargetAssets(0), 100 ether);
+        _dai.setLiquidity(100 ether);   // Simulate adding liquidity
```
rockyfour, line 102: Mmm not very critical but this returning to liquidity of 100 each time just to change it the next line seems a bit redundant.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -0,0 +1,158 @@
...
+        _dai.setLiquidity(40 ether);
+        plan.file("buffer", 100 ether);
+        assertEq(plan.getTargetAssets(0), 60 ether);
+    }
+
+    function test_increase_liquidity() public {
+        plan.file("buffer", 100 ether);
+        assertEq(plan.getTargetAssets(20 ether), 120 ether);
+    }
+
```
rockyfour, line 98: You don't have to change but I'd probably write a test for one case at a time and just try to cover all of them. I guess that's:

liquidity < buffer
liquidity = buffer
liquidity > buffer, partial unwind
liquidity > buffer, full unwind
buffer = 0

Just think this is easier on the eye than the simulation approach and makes sure we cover all the cases.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
rockyfour: In general, it's probably better to have an unauthed failure test for each specific authed function (so here it would be something like `test_disable_unauthed`).

The reasoning is that if we have more authed functions (apart from the standard ones which are tested in the base) we'll need specific tests anyway.

src/tests/pools/D3MAaveV3NoSupplyCapTypePool.t.sol
```diff
@@ -0,0 +1,329 @@
...
+
+    constructor(uint256 decimals_) D3MTestGem(decimals_) {
+        rewardsClaimer = address(new FakeRewardsClaimer());
+    }
+
+    function scaledBalanceOf(address who) external view returns (uint256) {
+        return balanceOf[who];
+    }
+
+    function setScaledTotalSupply(uint256 amount) external {
```
rockyfour, line 41: unused now

src/tests/pools/D3MAaveV3NoSupplyCapTypePool.t.sol
```diff
@@ -0,0 +1,329 @@
...
+
+import { D3MAaveV3NoSupplyCapTypePool, PoolLike } from "../../pools/D3MAaveV3NoSupplyCapTypePool.sol";
+
+interface RewardsClaimerLike {
+    function getRewardsBalance(address[] calldata assets, address user) external view returns (uint256);
+}
+
+contract AToken is D3MTestGem {
+    address public rewardsClaimer;
+    uint256 public scaledTotalSupply;
```
rockyfour, line 31: unused now

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
hexonaut: Yeah I know, but it separates the high-level actions such as "D3M adds liquidity", "User borrows", etc. Any issue if I leave it?

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
hexonaut: Sure thing. This is better if there is a bunch of authed functions anyways.

src/tests/pools/D3MAaveV3NoSupplyCapTypePool.t.sol
hexonaut: done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/55b326a9e5fb73a6b8451ee5dddc0eb9b899ddfa

src/tests/pools/D3MAaveV3NoSupplyCapTypePool.t.sol
hexonaut: done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/55b326a9e5fb73a6b8451ee5dddc0eb9b899ddfa

src/pools/D3MAaveV3NoSupplyCapTypePool.sol
hexonaut: done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/55b326a9e5fb73a6b8451ee5dddc0eb9b899ddfa

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
hexonaut: done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/4aa69542708c2664c78e2fe071c85d734db5ce05

src/plans/D3MAaveTypeBufferPlan.sol
hexonaut: Going to leave as-is then unles objection.

gbalabasquer: Sounds good. Resolving.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -0,0 +1,187 @@
...
+    function test_disable_unauthed() public {
+        plan.file("buffer", 1);
+        plan.deny(address(this));
+
+        vm.expectRevert("D3MAaveTypeBufferPlan/not-authorized");
+        plan.disable();
+    }
+
+    function test_implements_getTargetAssets() public override {
```
gbalabasquer, line 87: This test does the exact same thing than the below one. I guess we still need this one to overwrite the base, but then, should we remove the prev one or make this one even more trivial or just different.
Maybe just a `return true`

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -0,0 +1,187 @@
...
+        assertEq(plan.getTargetAssets(0), 60 ether);
+    }
+
+    function test_liquidity_equals_buffer() public {
+        _dai.setLiquidity(100 ether);
+        plan.file("buffer", 100 ether);
+        assertEq(plan.getTargetAssets(0), 0);
+    }
+
+    function test_liquidity_greater_than_buffer_full_unwind() public {
```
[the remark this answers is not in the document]
hexonaut: Oops yes you're right. Switched them.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
hexonaut: Simplified it.

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,414 @@
...
+    ATokenLike adai;
+    TreasuryLike treasury;
+    TreasuryAdminLike treasuryAdmin;
+    uint256 buffer;
+
+    D3MAaveTypeBufferPlan plan;
+    D3MAaveV3NoSupplyCapTypePool pool;
+
+    function setUp() public {
+        baseInit("template-spark");
```
gbalabasquer, line 135: Using the template input for running the tests seems weird. I don't know a better solution though, so this is probably fine for now but we should think more about this.

hexonaut: Ya agree. I didn't like it either, but couldn't think of a better solution.

gbalabasquer: Btw I think using the same `ilk` than then we will have in production will make the tests to fail once this is in production.

gbalabasquer: Actually looking at it a bit more, we can probably hardcode values here in the tests for the config and we can pass the necessary ones to `baseInit`.
I'd go for that option than using a file for this.

hexonaut: Yeah ok I like this better. done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/8c82a0c3f0a5e7f581e7a3c7a5d66b8adfdb46a2

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,414 @@
...
+
+        sparkPool = PoolLike(config.readAddress(".lendingPool"));
+        configurator = sparkPool.ADDRESSES_PROVIDER().getPoolConfigurator();
```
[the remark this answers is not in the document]
hexonaut: done https://github.com/makerdao/dss-direct-deposit/pull/102/commits/8c82a0c3f0a5e7f581e7a3c7a5d66b8adfdb46a2

src/tests/integration/SparkLend.t.sol
gbalabasquer: Yeah much better!

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,401 @@
...
+    function setUp() public {
+        baseInit();
+
+        sparkPool = PoolLike(0xC13e21B648A5Ee794902342038FF3aDAB66BE987);
+        daiInterestRateStrategy = DaiInterestRateStrategyLike(getInterestRateStrategy(address(dai)));
+        adai = ATokenLike(0x4DEDf26112B3Ec8eC46e7E31EA5e123490B05B8B);
+        treasury = TreasuryLike(adai.RESERVE_TREASURY_ADDRESS());
+        treasuryAdmin = treasury.getFundsAdmin();
+        buffer = 5_000_000 * WAD;
+        assertGt(buffer, 0);
```
gbalabasquer, line 132:
```suggestion
```

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        assertEq(vat.sin(address(vow)), sin + art * RAY);
+        assertEq(vat.dai(address(vow)), vowDai);
+
+        // Some time later the pool gets some liquidity
+        adjustLiquidity(-borrowAmount);
+
+        // Close out the remainder of the position
+        uint256 balance = getLPTokenBalanceInAssets(address(pool));
+        assertGe(balance, art);
+        hub.exec(ilk);
```
gbalabasquer, line 203:
```suggestion
        hub.exec(ilk);
        assertEq(getLPTokenBalanceInAssets(address(pool)), 0);
```

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+
+    function test_collect_interest() public {
+        adjustDebt(standardDebtSize);
+
+        generateInterest();
+
+        uint256 vowDai = vat.dai(address(vow));
+        hub.exec(ilk);
+
+        assertGt(vat.dai(address(vow)) - vowDai, 0);
```
gbalabasquer, line 246:
```suggestion
        assertGt(vat.dai(address(vow)), vowDai);
```
Just to be annoying :)

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,939 @@
...
+        // Cage the system to trigger only unwinds
+        vm.prank(admin); hub.cage(ilk);
+        hub.exec(ilk);
+
+        uint256 assets = getLPTokenBalanceInAssets(address(pool));
+        // All the fees are accrued but what can't be withdrawn is added up to the original ink and art debt
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(pool));
+        assertEq(ink, art);
+        assertRoundingEq(ink, assets);
+        assertGt(getLPTokenBalanceInAssets(address(pool)), 0);
```
gbalabasquer, line 286:
```suggestion
        assertGt(assets, 0);
```

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,939 @@
...
+        vm.prank(admin); hub.cage(ilk);
+        hub.exec(ilk);
+
+        uint256 assets = getLPTokenBalanceInAssets(address(pool));
+        // All the fees are accrued but what can't be withdrawn is added up to the original ink and art debt
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(pool));
+        assertEq(ink, art);
+        assertRoundingEq(ink, assets);
+        assertGt(assets, 0);
+        assertRoundingEq(ink, pink + feesAccrued - currentLiquidity);
```
gbalabasquer, line 287:
```suggestion
        uint256 amountToBorrow = currentLiquidity;
        adjustLiquidity(-int256(amountToBorrow));

        // Accumulate interest
        generateInterest();

        uint256 feesAccrued = getLPTokenBalanceInAssets(address(pool)) - pAssets;
        currentLiquidity = getLiquidity();

        assertGt(feesAccrued, 0);
        assertEq(pink, currentLiquidity);
        assertGt(pink + feesAccrued, currentLiquidity);

        // Cage the system to trigger only unwinds
        vm.prank(admin); hub.cage(ilk);
        hub.exec(ilk);

        uint256 assets = getLPTokenBalanceInAssets(address(pool));
        // All the fees are accrued but what can't be withdrawn is added up to the original ink and art debt
        (uint256 ink, uint256 art) = vat.urns(ilk, address(pool));
        assertEq(ink, art);
        assertRoundingEq(ink, assets);
        assertGt(assets, 0);
        assertRoundingEq(ink, feesAccrued);
```

gbalabasquer: Basically the change is to remove `amountSupplied` as it is the same than `pink` and also simplify the last assert which equals to `feesAccrued ` no need to sum and sub the same value (and it was actually pretty confusing).

gbalabasquer: The plan is at some point we migrate the V2 tests to use this base, right?

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        (, , uint256 tau, , ) = hub.ilks(ilk);
+        vm.warp(block.timestamp + tau);
+
+        hub.cull(ilk);
+
+        assertRevert(address(hub), abi.encodeWithSignature("uncull(bytes32)", ilk), "D3MHub/no-uncull-normal-operation");
+    }
+
+    function test_cage_exit() public {
+        adjustDebt(100 ether);
```
gbalabasquer, line 639: So maybe the main difference of this test vs the one in V2 is that the `totalArt` will match the amount of gems we are giving to the user. Maybe I'd just increase the buffer to something else, so we can prove the checks still hold if we are redeeming a portion (which is a more realistic case).

hexonaut: fixed

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(pool));
+        assertEq(ink, pink);
+        assertEq(art, part - 10 * WAD);
+        assertEq(ink - art, 10 * WAD);
+        assertEq(vat.gem(ilk, address(pool)), gemBefore);
+        assertEq(vat.vice(), viceBefore);
+        assertEq(vat.sin(address(vow)), sinBefore);
+        assertEq(vat.dai(address(vow)), vowDaiBefore);
+        assertRoundingEq(getLiquidity(), liquidityBalanceBefore);
+        assertApproxEqAbs(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore, 2 * roundingTolerance); // rounding may affect twice
```
gbalabasquer, line 844:
```suggestion
        assertEq(getLiquidity(), liquidityBalanceBefore);
        assertEq(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore, 2 * roundingTolerance);
```
Nothing was executed  that could actually change these values to have to round anything.

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        (uint256 ink, uint256 art) = vat.urns(ilk, address(pool));
+        assertEq(ink, pink);
+        assertEq(art, part - 10 * WAD);
+        assertEq(ink - art, 10 * WAD);
+        assertEq(vat.gem(ilk, address(pool)), gemBefore);
+        assertEq(vat.vice(), viceBefore);
+        assertEq(vat.sin(address(vow)), sinBefore);
+        assertEq(vat.dai(address(vow)), vowDaiBefore);
+        assertRoundingEq(getLiquidity(), liquidityBalanceBefore);
+        assertRoundingEq(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore);
```
gbalabasquer, line 895:
```suggestion
        assertEq(getLiquidity(), liquidityBalanceBefore);
        assertEq(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore);
```

gbalabasquer: Same here.

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        (ink, art) = vat.urns(ilk, address(pool));
+        assertRoundingEq(ink, pink);
+        assertRoundingEq(art, part);
+        assertEq(ink, art);
+        assertEq(vat.gem(ilk, address(pool)), gemBefore);
+        assertEq(vat.vice(), viceBefore);
+        assertEq(vat.sin(address(vow)), sinBefore);
+        assertApproxEqAbs(vat.dai(address(vow)), vowDaiBefore + 10 * RAD, RAY * roundingTolerance);
+        assertRoundingEq(getLiquidity(), liquidityBalanceBefore);
+        assertApproxEqAbs(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore, 2 * roundingTolerance); // rounding may affect twice
```
gbalabasquer, line 908:
```suggestion
        assertEq(getLiquidity(), liquidityBalanceBefore);
        assertEq(getLPTokenBalanceInAssets(address(pool)), assetsBalanceBefore);
```

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,400 @@
...
+
+        // User repays half their debt
+        assertEq(getDebt(), buffer + buffer / 2);
+        sparkPool.repay(address(dai), buffer / 4, 2, address(this));
+        assertEq(getDebt(), buffer + buffer / 2);
+
+        hub.exec(ilk);
+
+        assertEq(getDebt(), buffer + buffer / 2 - buffer / 4, "should be back down to 1.25x the buffer");
+    }
```
gbalabasquer, line 354:
```suggestion
    function test_simple_wind_unwind() public {
        setLiquidityToZero();

        assertEq(getDebt(), 0);

        hub.exec(ilk);
        assertEq(getDebt(), buffer, "should wind up to the buffer");

        // User borrows half the debt injected by the D3M
        sparkPool.borrow(address(dai), buffer / 2, 2, 0, address(this));
        assertEq(getDebt(), buffer);

        hub.exec(ilk);
        assertEq(getDebt(), buffer + buffer / 2, "should have 1.5x the buffer in debt");

        // User repays half their debt
        sparkPool.repay(address(dai), buffer / 4, 2, address(this));
        assertEq(getDebt(), buffer + buffer / 2);

        hub.exec(ilk);
        assertEq(getDebt(), buffer + buffer / 2 - buffer / 4, "should be back down to 1.25x the buffer");
    }
```

gbalabasquer: Replace the 3 tests just for one.

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,400 @@
...
+        return true;
+    }
+
+    function getTotalAssets(address asset) internal view returns (uint256) {
+        // Assets = DAI Liquidity + Total Debt
+        PoolLike.ReserveData memory data = sparkPool.getReserveData(asset);
+        return dai.balanceOf(address(adai)) + ATokenLike(data.variableDebtTokenAddress).totalSupply() + ATokenLike(data.stableDebtTokenAddress).totalSupply();
+    }
+
+    function getTotalLiabilities(address asset) internal view returns (uint256) {
```
gbalabasquer, line 304: What's the point of this wrapper function?

gbalabasquer: I think we can do:
```

    function getTotalLiabilities(address asset) internal view returns (uint256) {
        // Liabilities = spDAI Supply + Amount Accrued to Treasury
        PoolLike.ReserveData memory data = sparkPool.getReserveData(asset);
        return _divup((adai.scaledTotalSupply() + uint256(data.accruedToTreasury)) * data.liquidityIndex, RAY);
    }
```
and remove `getSupplyUsed`

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,400 @@
...
+        return art;
+    }
+
+    function _divup(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        unchecked {
+            z = x != 0 ? ((x - 1) / y) + 1 : 0;
+        }
+    }
+
+    function getSupplyUsed(address asset) internal view returns (uint256) {
```
rockyfour, line 272: Isn't this calculation just the total supply? I mean I see it used here where I'd expect the total supply to be used - https://github.com/aave/aave-v3-core/blob/94e571f3a7465201881a59555314cd550ccfda57/contracts/protocol/libraries/logic/ValidationLogic.sol#L74

src/tests/integration/SparkLend.t.sol
rockyfour, line 304: Can you explain what you mean by total liabilities? Is it the amount of DAI the pool owns to all lenders and to the treasury?

script/D3MDeploy.s.sol
```diff
@@ -116,6 +125,16 @@ contract D3MDeployScript is Script {
...
             }
+        } else if (planType.eq("liquidity-buffer")) {
+            if (poolType.eq("aave-v2") || poolType.eq("aave-v3-no-supply-cap")) {
+                d3m.plan = D3MDeploy.deployAaveBufferPlan(
+                    msg.sender,
+                    admin,
+                    config.readAddress(".adai")
+                );
+            } else {
+                revert("Invalid pool type for rate target plan type");
```
rockyfour, line 136:
```suggestion
                revert("Invalid pool type for liquidity buffer plan type");
```

script/D3MInit.s.sol
```diff
@@ -150,6 +152,19 @@ contract D3MInitScript is Script {
...
+                D3MAaveBufferPlanConfig memory aaveCfg = D3MAaveBufferPlanConfig({
+                    buffer: config.readUint(".buffer") * WAD,
+                    adai: D3MAavePoolLike(d3m.pool).adai()
+                });
+                D3MInit.initAaveBufferPlan(
+                    d3m,
+                    aaveCfg
+                );
+            } else {
+                revert("Invalid pool type for rate target plan type");
```
rockyfour, line 166:
```suggestion
                revert("Invalid pool type for liquidity buffer plan type");
```

script/D3MDeploy.s.sol
```diff
@@ -86,6 +86,15 @@ contract D3MDeployScript is Script {
                 address(dss.dai),
                 config.readAddress(".lendingPool")
             );
+        } else if (poolType.eq("aave-v3-no-supply-cap")) {
+            d3m.pool = D3MDeploy.deployAaveV3NoSupplyCapTypePool(
+                msg.sender,
```
rockyfour, line 91: A general suggestion for these scripts is to use named arguments, which can reduce the chance of mistakes (like confusing deployer and admin addresses for example).
Can maybe do it as a separate PR once this is merged.

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,400 @@
...
+        // Donate the excess liabilities back to the pool
+        // This will burn the liabilities while keeping the assets the same
+        vm.prank(admin); treasuryAdmin.transfer(address(treasury), address(adai), address(this), delta);
+        sparkPool.withdraw(address(dai), delta, address(adai));
+
+        assets = getTotalAssets(address(dai)) + 1;  // In case of rounding error we +1
+        liabilities = getTotalLiabilities(address(dai));
+        assertGe(assets, liabilities, "assets should be greater than or equal to liabilities");
+    }
+}
```
rockyfour, line 400: FYI I think `test_collect_stkaave` and `test_collect_stkaave_king_not_set` do not have counterparts here.

src/tests/integration/SparkLend.t.sol
hexonaut: Supply used is the same as total liabilities. I probably should have just merged the two, but supply used isn't even being used anymore, so I'll just get rid of it like you say.

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,938 @@
...
+        adjustLiquidity(int256(amountToBorrow));
+
+        // Rest of the liquidity can be withdrawn
+        hub.exec(ilk);
+        vow.heal(_min(vat.sin(address(vow)), vat.dai(address(vow))));
+        assertEq(vat.gem(ilk, address(end)), 0);
+        assertEq(vat.sin(address(vow)), 0);
+        assertGe(vat.dai(address(vow)), prevDai); // As also probably accrues interest
+    }
+
```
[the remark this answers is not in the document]
hexonaut: What do you mean? It's in the base. Line 334.

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,400 @@
...
+        DSTokenAbstract weth = DSTokenAbstract(dss.getIlk("ETH", "A").gem);
+        weth.setBalance(address(this), amt);
+        weth.approve(address(sparkPool), type(uint256).max);
+        dai.approve(address(sparkPool), type(uint256).max);
+        sparkPool.supply(address(weth), amt, address(this), 0);
+
+        assertGt(getDebtCeiling(), 0);
+
+        // Recompute the dai interest rate strategy to ensure the new line is taken into account
+        daiInterestRateStrategy.recompute();
```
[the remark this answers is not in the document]
hexonaut: This is a good point. I'll update it in the other repo.

src/tests/integration/SparkLend.t.sol
hexonaut: Yes exactly.

```
Total Assets = DAI in aDAI Contract (Idle Liquidity) + Stable Debt + Variable Debt
Total Liabilities = aDAI Supply (Users Lending) + Amount Owed to Treasury
```

src/tests/integration/SparkLend.t.sol
```diff
@@ -0,0 +1,504 @@
...
+
+        // Withdrawing won't work because no available DAI
+        vm.expectRevert();
+        sparkPool.withdraw(address(dai), delta, address(adai));
+
+        // Flash loan to close out the liabilities
+        flashLender.flashLoan(this, address(dai), delta, "");
+
+        // Someone puts the DAI back into the pool
+        dai.setBalance(address(adai), dai.balanceOf(address(adai)) + poolBalance);
```
rockyfour, line 482: Can we do it with actually pool deposit? it is part of the sequence we want to check.
Also we should use `delta` right?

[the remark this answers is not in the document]
gbalabasquer: Minor change to make it less artificial.

gbalabasquer (approved):
Giving my approval based on the same exact conditions of @rockyfour 's existing one.
D3M code per se looks good.

hexonaut:
Please note we are deploying with the reserve factor set to 0% and the performanceBonus set to 0. All funds will go to suppliers, and the accounting to the SubDAO will be handled off-chain for now.

PR #104, opened 2023-03-17
Add PSM Support for Investment Vehicles

hexonaut:
This will merge the PSM into the D3M design by pre-minting the DAI into the swap pool. The pool will be designed to target a particular desired exposure to an asset and use negative fees to pay arbitragers to fill the pool.

Also included is a first version of an ALM controller to ensure liquidity is always available in the USDC PSM if needed.


hexonaut:
I'm going to add some more tests, but this is getting close to being finished for review.

PR #105, opened 2023-03-17
Kinked Swap Pool

hexonaut:
NOTE: This is no longer needed. Leaving in case someone needs a kinked pool in the future. Pools that will be used have been merged into ALM Controller PR: https://github.com/makerdao/dss-direct-deposit/pull/104

psm-2 0-design

This will merge the PSM into the D3M design by pre-minting the DAI into the swap pool. The pool will be designed to target a particular desired exposure to an asset and use negative fees to pay arbitragers to fill the pool.

Contains the following new features beyond the original PSM:

 * DAI is pre-minted into the swap contract via D3M. This will save a lot of gas and is standard for D3M interfaces.

psm-2 0-oracles

 * Swaps are performed at an Oracle price instead of 1:1 with a buy and sell oracle.
 * No more dealing with fees inside the PSM. Using standard `assetBalance()` to return market valuation and the `D3MHub` will compare assets vs liabilities to determine how much to send to surplus buffer (or take back).

psm-2 0-ratio-targeting

 * `ratio` parameter is added to set a desired gem/dai ratio. Additionally `tin` and `tout` are split into two versions (one for below the desired gem/dai ratio and one above). This serves a few use cases:

1. Tokens that are being used as investments we want ways for arbitragers to wind them to desired levels by using negative fees.
2. Having an extra buffer allows for payments in the token itself. For example, Gemini can just send GUSD tokens to the swap contract to automatically transfer the value to the surplus buffer via D3M asset/liability accounting. Arbitragers would then be able to periodically clear it out back to the target ratio level (where the `tout` fee would switch to a higher value to prevent full unwind). In the GUSD case this would be like keeping 490m GUSD and 10m DAI so that interest can be paid periodically without going over the 500m debt ceiling.
3. For USDC where we may want to increase the redemption price a `ratio` can be set to some low amount say 40% where if the gem/dai ratio drops below this then the `tout` increases to a larger number for market price discovery.

Uses:

 * Backed Finance bTokens
 * Ondo OUSG
 * GUSD

### DIFF
--- a/src/pools/D3MLinearFeeSwapPool.sol
+++ b/src/pools/D3MLinearFeeSwapPool.sol
@@ -112,7 +112,7 @@ contract D3MLinearFeeSwapPool is D3MSwapPool {
         uint256 totalBalance = daiBalance + gemBalance;
         if (totalBalance > 0) {
             // Please note the fee deduction is not included in the new total dai+gem balance to drastically simplify the calculation
-            fee = (_feeData.tout1 * daiBalance + _feeData.tout2 * gemBalance - (_feeData.tout1 + _feeData.tout2) * daiAmt / 2) / totalBalance;
+            fee = (_feeData.tout2 * daiBalance + _feeData.tout1 * gemBalance - (_feeData.tout1 + _feeData.tout2) * daiAmt / 2) / totalBalance;
         }
         uint256 gemValue = daiAmt * fee / BPS;
         gemAmt = gemValue * WAD / (GEM_CONVERSION_FACTOR * pipValue);


chechu:
Great PR, thanks for sharing it. I was trying to understand the Oracles graph, and I had some doubts:

(I will ignore fees in every case, assuming they don't exist)
* top-left. Applicable when the user sends DAI to the contract and receives the Gem (i.e. USDC), right? My understanding is: if the USDC market price is $0.5, and the user sends 1 DAI, the PSM will send to the user 2 USDC. But if the market price of USDC is $1.5 and the user sends 1 DAI, the PSM will send to the user 1 USDC. Is that correct? 
* top-right. Applicable when the user sends Gem (i.e. USDC) to the contract and receives DAI, right? My understanding is: if the USDC market price is $1.5, and the user sends 1 USDC, the PSM will send to the user 1.5 DAI. But if the market price of USDC is $0.5 and the user sends 1 USDC, the PSM will send to the user 1 DAI. Is that correct? 
* in the bottom graphs, what NAV means?

Thanks again ❤️

hexonaut:
The directions are the opposite. It's from the protocol's perspective on how much to pay for buying a gem (say USDC).

NAV = Net Asset Value, the book value of the assets in this example.

PR #107, opened 2023-03-22
Major Test Refactor

hexonaut:
This PR is doing test cleanup to prepare for the horizontal scaling of the D3M system. If left unaddressed this will increase the tech debt down the line.

Summary of Changes:

1. D3MHub unit tests are very specific to the Aave Pool/Plan doing things like setting `bar` in the plan. There were also a lot of tests inside `D3MHubTest` testing the functionality of the plan/pool mocks which didn't really add any value. These holistic tests should be moved to a full integration test suite. I've stripped out everything that is being tested unnecessarily. These should also be converted into full unit tests by mocking all the dss contracts instead of using a mainnet fork.
1. Drop this https://github.com/makerdao/dss-direct-deposit/blob/master/src/tests/interfaces/interfaces.sol and use dss-interfaces.
1. Use dss-test primitives. Things like: `checkAuth()`, `checkFile()`, etc. Give tokens functions can all be converted to either the GodMode version or forge-std's implement: https://github.com/makerdao/dss-direct-deposit/blob/master/src/tests/pools/D3MPoolBase.t.sol#L174 . Tests [like this](https://github.com/makerdao/dss-direct-deposit/blob/master/src/tests/pools/D3MPoolBase.t.sol#L214) should be eliminated in favour of `checkAuth()`.
1. Cleaning up the Pool and Plan base contracts. There was a lot of repetitive casting and unnecessary tests. As well some common behavior such as `pool.exit()` is not being tested in the abstract parent. This was the most important changes as these will be extended for all future Pool/Plans. 

Overall this removes around 1200 lines of what I view as unnecessary code.

Still todo in another PR (not blocking new D3M dev): Several of the "unit" tests are loading in existing contracts from mainnet such as here: https://github.com/makerdao/dss-direct-deposit/blob/master/src/tests/plans/D3MAavePlan.t.sol#L62 . All these contracts should be mocked so there is no requirement to fork mainnet.


test.sh
```diff
@@ -22,25 +22,25 @@ do
 done
 
 if [[ -z "$MATCH" && -z "$BLOCK" && -z "$MATCH_TEST" && -z "$MATCH_CONTRACT" ]]; then
-    forge test --use "$SOLC" --fork-url "$ETH_RPC_URL" -vv --force
+    forge test --use "$SOLC" --fork-url "$ETH_RPC_URL" -vv
```
gbalabasquer, line 25: This is already corrected on `master` we should probably get it merged here to avoid this change again.

src/tests/D3MMom.t.sol
```diff
@@ -16,50 +16,43 @@
...
     function test_disable_no_auth() public {
-        d3mMom.setOwner(address(0));
-        assertEq(d3mMom.authority(), address(0));
-        assertEq(d3mMom.owner(), address(0));
+        mom.setOwner(address(0));
+        assertEq(mom.authority(), address(0));
+        assertEq(mom.owner(), address(0));
 
-        assertRevert(address(d3mMom), abi.encodeWithSignature("disable(address)", address(d3mTestPlan)), "D3MMom/not-authorized");
+        assertRevert(address(mom), abi.encodeWithSignature("disable(address)", plan), "D3MMom/not-authorized");
```
gbalabasquer, line 55: Should we start removing the `assertRevert` in favor the native Foundry `expectRevert`?

test.sh
hexonaut: merged.

src/tests/stubs/D3MTestPlan.sol
```diff
@@ -1,95 +0,0 @@
-// SPDX-FileCopyrightText: © 2022 Dai Foundation <www.daifoundation.org>
```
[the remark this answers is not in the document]
hexonaut: readded. feel free to make modifications

src/tests/D3MHub.t.sol
```diff
@@ -16,704 +16,646 @@
...
-        assertTrue(d3mTestPool.postDebt());
+        assertTrue(pool.preDebt());
+        assertTrue(pool.postDebt());
     }
 
     function test_wind_limited_by_maxDeposit() public {
-        _windSystem(); // winds to 50 * WAD
-        d3mTestPlan.file("targetAssets", 75 * WAD);
-        d3mTestPool.file("maxDepositAmount", 5 * WAD);
+        plan.setTargetAssets(50 * WAD);
```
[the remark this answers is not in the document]
hexonaut: Just a mistake. Fixed.

src/tests/D3MHub.t.sol
```diff
@@ -881,95 +808,91 @@ contract D3MHubTest is DssTest {
...
+        hub.exec(otherIlk);
 
         (ink, art) = vat.urns(otherIlk, address(otherPool));
         assertEq(ink, 5 * WAD);
         assertEq(art, 5 * WAD);
         assertTrue(otherPool.preDebt());
         assertTrue(otherPool.postDebt());
-        otherPool.file("preDebt", false); // reset preDebt
-        otherPool.file("postDebt", false); // reset postDebt
+        otherPool.resetPrePostDebt();
```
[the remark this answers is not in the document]
hexonaut: This one actually makes sense to test that it was tripped in the `hub`.

src/tests/D3MHub.t.sol
```diff
@@ -16,704 +16,647 @@
...
+
+    function maxWithdraw() external view override returns (uint256) {
+        return dai.balanceOf(address(this));
+    }
+
+    function redeemable() external view returns (address) {
+        return address(gem);
+    }
+
+    function decreaseIdleLiquidity(uint256 amount) external {
```
gbalabasquer, line 89: Can I get a TLDR of what exactly means this increasing/decreasing `DAI` supply against the `gem`? I'm a bit confused why it is being used this way.
Is it to represent how much is lent? 
If that's the case I'm a bit confused why when it is used says this "liquidity before simulating other user's withdraw", shouldn't be there term "borrow" instead of "withdraw"?

src/tests/D3MHub.t.sol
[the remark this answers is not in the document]
gbalabasquer: Hmm but that one doesn't test anything, just reset the two values state, and the function is the test itself which ends there.
I mean it makes sense at the end of `_windSystem` so you can have the values reset for the test itself that is calling this `internal`. However at the end of the test don't see the usage.

src/tests/D3MHub.t.sol
```diff
@@ -881,95 +808,91 @@ contract D3MHubTest is DssTest {
...
         // Make sure pre/post functions get called
-        assertTrue(d3mTestPool.preDebt());
-        assertTrue(d3mTestPool.postDebt());
+        assertTrue(pool.preDebt());
+        assertTrue(pool.postDebt());
     }
 
     function test_exit() public {
         _windSystem();
+        pool.decreaseIdleLiquidity(50 * WAD);
```
gbalabasquer, line 888: I see this is being used here to have token availability in the pool as in the Mock we are not minting anymore. It feels a bit hacky though, and I think this comes also related to my previous comment of what increase/decreaseIdleLiquidity. Wondering if there is a bit better way to handle that, maybe more similar to how the mock was before. Did you see any major problem with that model?

src/tests/D3MHub.t.sol
```diff
@@ -978,285 +901,214 @@ contract D3MHubTest is DssTest {
...
-        d3mHub.cage(ilk);
+        hub.cage(ilk);
 
-        d3mHub.cull(ilk);
+        hub.cull(ilk);
 
-        assertRevert(address(d3mHub), abi.encodeWithSignature("uncull(bytes32)", ilk), "D3MHub/no-uncull-normal-operation");
-    }
-
-    function test_quit_culled() public {
```
gbalabasquer, line 1183: I know these two are not actually test anything in the `Hub` itself, but they were kind of important to prove we can move the position. Do you think we can actually keep them? Maybe removing the pool part.

src/tests/D3MHub.t.sol
```diff
@@ -1293,315 +1144,28 @@ contract D3MHubTest is DssTest {
...
-
-        // New Plan should determine the pool position
-        (uint256 ink, uint256 art) = vat.urns(ilk, address(d3mTestPool));
-        assertEq(ink, 100 * WAD);
-        assertEq(art, 100 * WAD);
-        assertTrue(d3mTestPool.preDebt());
-        assertTrue(d3mTestPool.postDebt());
-    }
-
-    function test_hub_upgrade_same_d3ms() public {
```
gbalabasquer, line 1416: Is there a problem to keep this one?

src/tests/D3MHub.t.sol
```diff
@@ -1293,315 +1144,28 @@ contract D3MHubTest is DssTest {
...
-        newHub.exec(ilk);
-
-        (uint256 ink, uint256 art) = vat.urns(ilk, address(d3mTestPool));
-        assertEq(ink, 100 * WAD);
-        assertEq(art, 100 * WAD);
-        assertTrue(d3mTestPool.preDebt());
-        assertTrue(d3mTestPool.postDebt());
-    }
-
-    function testFail_hub_upgrade_kills_old_hub() public {
```
gbalabasquer, line 1454: Same here.

src/tests/D3MHub.t.sol
```diff
@@ -1611,25 +1175,25 @@ contract D3MHubTest is DssTest {
         vat.file(ilk, "line", 60 * RAD);
 
         // Simulate a loss event by removing the share tokens
```
gbalabasquer, line 1177: Hmm for sure this comment is misleading now. But another point to make me wonder if we can fallback to something more similar it was before.

src/tests/D3MHub.t.sol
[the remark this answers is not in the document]
hexonaut: Ah yes I see what you're saying. Ok fixed.

src/tests/D3MHub.t.sol
hexonaut: Yeah so the fundamental thing is that the pool has some notion of "idle liquidity" which is DAI that is immediately available. The mock is just using a `gem` to represent some IOU asset for the DAI so the proper `assetBalance()` can be calculated. Really this could have probably been further simplified to just a number in the mock instead of an ERC20.

For the second point borrow or withdraw is fine. Both result in the DAI sitting in the pool going away. In the borrow case another user has borrowed some of the previously idle DAI. In the withdraw case someone was previously lending and withdrew. Of course this is all in the context of Aave/Compound and shouldn't probably have that language in the comments. I'll remove the comment.

src/tests/D3MHub.t.sol
hexonaut: These tests should be re-added as integration tests imo. Having them here doesn't really show us anything.

src/tests/D3MHub.t.sol
hexonaut: I think this is a more minimal version of the mock. DAI comes in via `pool.deposit()` and later goes out to somewhere else to be replaced with another token. Could be aDAI, cDAI, but it's also valid for this to just be 1 USDC or something. Maybe the name of the function `decreaseIdleLiquidity()` is confusing?

src/tests/D3MHub.t.sol
hexonaut: Same as above, I think this should be an integration test. This as a unit test provides a false sense of coverage imo, but if you really want to readd I won't die on this hill.

src/tests/D3MHub.t.sol
gbalabasquer: I understand you want to make an abstraction, but then something like `test_exit` looks very weird. As why you need to decrease the liquidity in order to have the exit token. That should be there always regardless how much DAI there is.
I'm proposing this PR which sets the things a bit more similar than before but keeping the Mock pretty simple: https://github.com/makerdao/dss-direct-deposit/pull/108.
If you feel that this breaks the abstraction you want to get, then let's keep discussing.

src/tests/plans/D3MPlanBase.t.sol
```diff
@@ -16,117 +16,38 @@
...
-
-    function test_implements_getTargetAssets() public virtual {
-        uint256 result = D3MPlanBase(d3mTestPlan).getTargetAssets(2);
-
-        assertEq(result, 2);
-    }
-
-    function test_implements_active() public view {
-        D3MPlanBase(d3mTestPlan).active();
-    }
```
gbalabasquer, line 127: Shouldn't these two be part of the base? The `hub` requires them, a plan that doesn't implement them is wrong.

src/tests/D3MHub.t.sol
hexonaut: So at its core there is no requirement in `deposit()`/`withdraw()` to convert into another token right away. This is what happens in Aave/Compound, but for a PSM swap the pool can be filled with some mix of the gem and the dai as you can see here: https://github.com/makerdao/dss-direct-deposit/pull/105/files#diff-09fcb804207fcd0316dc1b3f8a584c9097581115543a1f2e561f0e4fc5ab6552R203-R209

The checks in the tests that the amount of gem balance has increased is specific to how Aave/Compound-style works and so that's why I'm trying to keep it as minimal as possible. Basically the pool is a black box where you stick some DAI in and then some random actions will occur in the future to change the mix between `dai` and the `redeemable()` token. This is how I'm modeling the increase/decrease idle dai functions. IE something happened, could be a swap, could be a borrow, could be a withdraw, doesn't matter. The point is the DAI available to immediately withdraw has dropped and some other asset has increased.

Does it clear it up?

src/tests/plans/D3MPlanBase.t.sol
hexonaut: That's enforced by inheriting the interface though.

gbalabasquer: Hmmm fair, I'd think we could still leave them, but I guess we are protected anyway so not strong opinion.

src/tests/pools/D3MPoolBase.t.sol
```diff
@@ -16,277 +16,106 @@
...
-    function test_cannot_file_hub_no_auth() public {
-        D3MPoolBase(d3mTestPool).deny(address(this));
+        assertEq(redeemableToken.balanceOf(TEST_ADDRESS), 0);
+        assertEq(ExitLike(address(pool)).exited(), 0);
 
-        assertRevert(address(d3mTestPool), abi.encodeWithSignature("file(bytes32,address)", bytes32("hub"), address(123)), string(abi.encodePacked(contractName, "/not-authorized")));
-    }
+        vm.prank(address(hub)); pool.exit(TEST_ADDRESS, 10 ether);  // Exit 10%
 
-    function test_cannot_file_hub_vat_caged() public {
```
gbalabasquer, line 253: I think this is one important to test at the base to make sure the implementation is safe for ES.

hexonaut: Oops yes accidentally deleted this. Re-added.

src/tests/D3MHub.t.sol
gbalabasquer: Let's wait for @rockyfour comment.

hexonaut: Sure. I will be adding this integration test in a future PR btw as Hub needs to have minor upgrades to support the Star Vaults.

src/tests/D3MHub.t.sol
gbalabasquer: After discussions, let's just update to this:
```suggestion
        // Simulate a loss event by removing the tokens
```

src/tests/D3MHub.t.sol
```diff
@@ -1638,298 +1201,266 @@ contract D3MHubTest is DssTest {
...
     function test_exec_fixInk_full_at_debt_ceiling() public {
         _windSystem();
         // interest is determined by the difference in gem balance to dai debt
         // by giving extra gems to the Join we simulate interest
-        _giveTokens(TokenLike(address(testGem)), 10 * WAD);
-        testGem.transfer(address(d3mTestPool), 10 * WAD); // Simulates 10 WAD of interest accumulated
-        assertEq(testGem.balanceOf(address(d3mTestPool)), 60 * WAD);
-        assertEq(d3mTestPool.maxWithdraw(), 50 * WAD);
+        address(testGem).setBalance(address(this), 10 * WAD);
+        testGem.transfer(address(pool), 10 * WAD); // Simulates 10 WAD of interest accumulated
```
gbalabasquer, line 1280: I'm noticing that sometimes we keep the old way of adding to address(this) and the transferring to the pool to make sure to not override existing balance but in others we set directly in the pool. I know for this PoolMock it would be the same as depositing doesn't add testGem anymore so we start with balance 0. However I'd believe we should standardize. Probably the less error prone is setBalance to the pool directly but adding up to the current balance.

gbalabasquer:
Ok I think I'm done with the review, just pending on the standardization of the `Hub` test and the comment from rocky about including or not those others.

src/tests/D3MHub.t.sol
hexonaut: Good point. I added a standard `accrueInterest` function to the pool.

src/tests/D3MHub.t.sol
rockyfour: I think it would be better to leave these in, at least until we have the same coverage in each integration test.
Apart from actually testing stuff, they serve as a "guide" on how these sequences look like.

(speaking about test_quit_culled, test_quit_not_culled, test_pool_upgrade_quit, test_plan_upgrade, test_hub_upgrade_same_d3ms, testFail_hub_upgrade_kills_old_hub, test_hub_upgrade_new_d3ms))

src/tests/D3MHub.t.sol
rockyfour: See my comment in the above discussion about `test_quit_culled`.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
```diff
@@ -84,36 +98,32 @@ contract D3MAaveTypeBufferPlanTest is D3MPlanBaseTest {
         plan.disable();
     }
 
-    function test_implements_getTargetAssets() public override {
```
rockyfour, line 87: I think if we remove this we should at least rename the similar-named test from `test_implements_getTargetAssets` to just `test_getTargetAssets`

rockyfour:
Made a relatively high-level review of the following:

* making sure no source contract file was changed (only testing related).
* making sure no undiscussed testing coverage was removed.
* following the GitHub discussion and making sure I'm fine with everything.
* making sure tests still pass

Under these assumptions, I will also ll be finished once the open issues are resolved.

src/tests/plans/D3MAaveTypeBufferPlan.t.sol
hexonaut: done

src/tests/D3MHub.t.sol
hexonaut: Ok, I'll re-add.

hexonaut: Ok it's been re-added.

hexonaut:
FYI I've fixed the IPool.quit() description as it should transfer out all assets not just the shares.

src/tests/D3MHub.t.sol
gbalabasquer: Resolving as it was re-added.

gbalabasquer (approved):
LGTM

rockyfour (approved):
LGTM

PR #112, opened 2023-05-09
chore: Update to use latest forge conventions

lucas-manuel:
- Removes the need for the ./test.sh script
- Allows for individual tests to use dedicated fork blocks if necessary (e.g., Spark after deployment)
- Allows for inline contract matching and verbosity specification

To run a given test: `forge t --mt <test>`
To debug with 4 verbosity (allows setUp examination on failure): `forge t --mt <test> -vvvv`


lucas-manuel:
Passing locally but failing in CI, investigating.

hexonaut:
Fixed the regular tests by updating rpc url to archive node. I think we can merge this as the Certora issue is unrelated and happening in master anyways.

PR #113, opened 2023-05-11
Fix requires and missing asserts

sunbreak1211:


sunbreak1211:
@hexonaut this is very trivial, so I'll just merge it to your branch. If there is anything missing we can change that there.

PR #115, opened 2024-03-14
Fix tests

MathisGD:
Aave changed the IRM of DAI on AaveV2 recently ([proposal](https://app.aave.com/governance/v3/proposal/?proposalId=3), [tx](https://etherscan.io/tx/0x32e0c2df50544c1756263bc7ea7eaabe569817bab825503f62fcf163ff1b5eda)), and it breaks the tests as its address is hardcoded in them. An other fix would have been to query the irm of DAI each time, but maybe you prefer keeping it like that.

nb: this [new irm](https://etherscan.io/address/0xab7f8ca781C736EB624A9E321cd043F8E6292C96#code) is not verified on Etherscan 🤔


hexonaut (approved):
LGTM

PR #116, opened 2024-03-15
D3M in 4626 vaults

MathisGD:


src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,163 @@
...
+        vat.hope(data);
+
+        emit File(what, data);
+    }
+
+    function urdClaim(address dst, address distributor, address reward, uint256 claimable, bytes32[] calldata proof) external payable {
+        require(dst != address(0), "D3M4626TypePool/zero-address");
+
+        uint256 amount = IUniversalRewardsDistributor(distributor).claim(address(this), reward, claimable, proof);
+        IERC20(reward).transfer(dst, amount);
```
hexonaut, line 133: This seems like users can permissionlessly steal the rewards of the d3m. All rewards should just accrue to king address same as the other pools: https://github.com/makerdao/dss-direct-deposit/blob/master/src/pools/D3MAaveV3NoSupplyCapTypePool.sol#L247

hexonaut: Actually in the interest of simplicity let's just remove the reward collection for now. We can upgrade the pool later if need be.

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,163 @@
...
+    /* CONSTANTS */
+
+    IERC20 public immutable dai;
+    IERC4626 public immutable vault;
+    VatLike public immutable vat;
+
+    /* STORAGE */
+
+    address public hub;
+    mapping(address => bool) public wards;
```
hexonaut, line 50: `wards` should be `uint256` instead of `bool` to match other interfaces

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    function test_cannot_file_king_no_auth() public {
+        pool.deny(address(this));
+        assertRevert(address(pool), abi.encodeWithSignature("file(bytes32,address)", bytes32("king"), address(123)), "D3M4626TypePool/not-authorized");
+    }
+
+    function test_cannot_file_king_vat_caged() public {
+        vat.cage();
+        assertRevert(address(pool), abi.encodeWithSignature("file(bytes32,address)", bytes32("king"), address(123)), "D3M4626TypePool/no-file-during-shutdown");
+    }
```
hexonaut, line 62: Drop the `king` we don't need rewards rn.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,167 @@
...
+    //     vm.mockCall(distributor, abi.encodeCall(IUniversalRewardsDistributor.claim, (address(pool), reward, claimable, proof)), abi.encode(1));
+    //     vm.mockCall(reward, abi.encodeCall(IERC20.transfer, (address(dst), 1)), abi.encode(true));
+
+    //     pool.urdClaim(dst, distributor, reward, claimable, proof);
+    // }
+
+    function test_urd_claim_no_dst(address distributor, address reward, uint256 claimable, bytes32[] calldata proof) public {
+        vm.expectRevert("D3M4626TypePool/zero-address");
+        pool.urdClaim(address(0), distributor, reward, claimable, proof);
+    }
```
hexonaut, line 105: Can also drop these as we are removing rewards support.

src/pools/D3M4626TypePool.sol
MathisGD: actually rewards are claimable by anybody (and will be transferred to the entitled account), so we could just have a onlyAuth skim function

src/pools/D3M4626TypePool.sol
MathisGD: Both are interchangeable, but if you prefer like that I change

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,150 @@
...
+    /// @inheritdoc ID3MPool
+    function withdraw(uint256 assets) external override onlyHub {
+        vault.withdraw(assets, msg.sender, address(this));
+    }
+
+    /// @inheritdoc ID3MPool
+    /// @dev prop = 100 ether exists 100%.
+    function exit(address dst, uint256 prop) external onlyHub {
+        uint256 shares = prop * vault.balanceOf(address(this)) / 100 ether;
+        vault.transfer(dst, shares);
```
MathisGD, line 92: should require true

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /* ONLY AUTHORIZED */
+
+    /// @inheritdoc ID3MPool
+    function quit(address dst) external auth {
+        require(vat.live() == 1, "D3M4626TypePool/no-quit-during-shutdown");
+        vault.transfer(dst, vault.balanceOf(address(this)));
```
MathisGD, line 100: should require true

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /// @inheritdoc ID3MPool
+    function withdraw(uint256 assets) external override onlyHub {
+        vault.withdraw(assets, msg.sender, address(this));
+    }
+
+    /// @inheritdoc ID3MPool
+    /// @dev prop = 100 ether exists 100%.
+    function exit(address dst, uint256 prop) external onlyHub {
```
hexonaut, line 90: This needs to be replaced with this:

```
function exit(address dst, uint256 wad) external override onlyHub {
    uint256 exited_ = exited;
    exited = exited_ + wad;
    uint256 amt = wad * vault.balanaceOf(address(this)) / (D3mHubLike(hub).end().Art(ilk) - exited_);
    require(vault.transfer(dst, amt), "D3M4626Pool/transfer-failed");
}
```

PR #117, opened 2024-03-15
MetaMorpho D3M

hexonaut:
D3M into MetaMorpho Vault with a multisig operator.


src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,94 @@
...
+import "./ID3MPlan.sol";
+
+/**
+ *  @title D3M Operator Plan
+ *  @notice An operator sets the desired target assets.
+ */
+contract D3MOperatorPlan is ID3MPlan {
+
+    mapping (address => uint256) public wards;
+    uint256                      public enabled;
```
MathisGD, line 28: why not?
```suggestion
    bool                         public active;
```

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    function setTargetAssets(uint256 value) external {
+        require(msg.sender == operator, "D3MOperatorPlan/not-authorized");
+
+        targetAssets = value;
+    }
+
+    function getTargetAssets(uint256) external override view returns (uint256) {
+        return targetAssets;
+    }
```
MathisGD, line 84: should it still return target asset if it's disabled?

hexonaut: Good catch. It will still unwind in the hub logic, but better to match the other plan's behaviour which return 0 when disabled.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "lib/dss-test"]
 	path = lib/dss-test
 	url = https://github.com/makerdao/dss-test
+[submodule "lib/solmate"]
```
sunbreak1211, line 4: I think it would be better to not include a third party git submodule and just copy the needed files for testing in a tests/utils or tests/helpers directory.

MathisGD: ok, will remove

MathisGD: but I think then I think that we need to keep the soulmate licence.
otherwise I can copy-paste the oz 4626

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,167 @@
...
+        wards[msg.sender] = 1;
+        emit Rely(msg.sender);
+
+        dai.approve(address(vault), type(uint256).max);
+    }
+
+    /* MODIFIERS */
+
+    modifier onlyHub() {
+        require(msg.sender == address(hub), "D3M4626TypePool/only-hub");
```
hexonaut, line 83: No need to cast `hub` to `address`.

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    /* ONLY AUTHORIZED */
+
+    /// @inheritdoc ID3MPool
+    function quit(address dst) external auth {
+        require(vat.live() == 1, "D3M4626TypePool/no-quit-during-shutdown");
+        require(vault.transfer(dst, vault.balanceOf(address(this))), "D3M4626TypePool/transfer-failed");
+    }
+
+    function rely(address usr) public auth {
```
hexonaut, line 120: Can be `external` instead of `public`.

hexonaut (commented):
Two minor issues. Looks good otherwise.

.gitmodules
sunbreak1211: Ok good, let's leave this change on hold until getting closer to merge the PR.

src/pools/D3M4626TypePool.sol
MathisGD:
```suggestion
        require(msg.sender == hub, "D3M4626TypePool/only-hub");
```

src/pools/D3M4626TypePool.sol
MathisGD:
```suggestion
    function rely(address usr) external auth {
```

hexonaut (commented):
Ok both contracts LGTM. Will do test review shortly.

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,167 @@
...
+interface D3mHubLike {
+    function vat() external view returns (address);
+    function end() external view returns (EndLike);
+}
+
+interface EndLike {
+    function Art(bytes32) external view returns (uint256);
+}
+
+contract D3M4626TypePool is ID3MPool {
```
sunbreak1211, line 37: I know we are trying to avoid styling changes, but considering this file is gonna be audited and then we won't want to touch it I'm mentioning this now:
Could we try to have a more general homogeneous order of things compared to the other pools?
Also the `wards` storage slot uses to be the 0 in almost all Maker contracts.

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (enabled == 0) return 0;
+
+        return targetAssets;
+    }
+
+    function active() public view override returns (bool) {
+        return enabled == 1;
+    }
+
+    function disable() external override auth {
```
sunbreak1211, line 92: The other plans have this `require` instead of the `auth` modifier:
```
require(wards[msg.sender] == 1 || !active(), ".../not-authorized");
```
Ofc it is an absurd for this plan, but it is also for the `D3MAaveTypeBufferPlan` (Spark one) and we can still find it there.
I think the reason was to avoid forgetting this condition for future plans where it actually has a purpose (as the rate base ones).
I don't have a strong opinion about keeping it or not but wanted to make the others aware of this.

src/pools/D3M4626TypePool.sol
hexonaut: Like the ordering? @MathisGD please rearrange to match the function / storage ordering of the other contracts if there is extra time.

src/plans/D3MOperatorPlan.sol
hexonaut: I would leave it as-is.

script/input/1/template-morpho.json
```diff
@@ -1,6 +1,6 @@
 {
     "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
-    "admin": "0x3300f198988e4C9C63F75dF86De36421f06af8c4",
+    "admin": "0xBE8E3e3618f7474F8cB1d074A26afFef007E98FB",
```
sunbreak1211, line 3: Ah I deleted my comment when I realized it was the Spark sub proxy. So it is going with the main pause proxy for now, right?

src/pools/D3M4626TypePool.sol
hexonaut: Fixed to match Spark pool to make diff easier

script/input/1/template-morpho.json
hexonaut: fixed, should be pause proxy for D3M yes

src/pools/D3M4626TypePool.sol
sunbreak1211: Looks good! Just a minor remaining thing, if we could switch onlyHub with auth. (don't hate me :))

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,153 @@
...
+        if (what == "hub") {
+            vat.nope(hub);
+            hub = data;
+            vat.hope(data);
+        } else revert("D3M4626TypePool/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    /// @inheritdoc ID3MPool
+    function deposit(uint256 wad) external override onlyHub {
```
sunbreak1211, line 105: It probably helps adding in a comment the link to the function:
https://github.com/morpho-org/metamorpho/blob/fcf3c41d9c113514c9af0bbf6298e88a1060b220/src/MetaMorpho.sol#L531

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,153 @@
...
+        vault.deposit(wad, address(this));
+    }
+
+    /// @inheritdoc ID3MPool
+    function withdraw(uint256 wad) external override onlyHub {
```
sunbreak1211, line 110: Same: https://github.com/morpho-org/metamorpho/blob/fcf3c41d9c113514c9af0bbf6298e88a1060b220/src/MetaMorpho.sol#L557

src/pools/D3M4626TypePool.sol
hexonaut: thought I did this. np, changed

src/pools/D3M4626TypePool.sol
hexonaut: done

src/pools/D3M4626TypePool.sol
hexonaut: done

src/pools/D3M4626TypePool.sol
```diff
@@ -0,0 +1,155 @@
...
+        require(hub_ != address(0), "D3M4626TypePool/zero-address");
+        require(dai_ != address(0), "D3M4626TypePool/zero-address");
+        require(vault_ != address(0), "D3M4626TypePool/zero-address");
+        require(IERC4626(vault_).asset() == dai_, "D3M4626TypePool/vault-asset-is-not-dai");
+
+        dai.approve(vault_, type(uint256).max);
+
+        hub = hub_;
+        vat = VatLike(D3mHubLike(hub).vat());
+        vat.hope(hub_);
```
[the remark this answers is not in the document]
hexonaut: fixed

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+    function file(bytes32 what, uint256 data) external auth {
+        if (what == "enabled") {
+            require(data <= 1, "D3MOperatorPlan/invalid-value");
+            enabled = data;
+        } else revert("D3MOperatorPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    function setTargetAssets(uint256 value) external {
+        require(msg.sender == operator, "D3MOperatorPlan/not-authorized");
```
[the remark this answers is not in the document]
hexonaut: done

src/tests/plans/D3MOperatorPlan.t.sol
```diff
@@ -0,0 +1,108 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity ^0.8.14;
+
+import { D3MPlanBaseTest } from "./D3MPlanBase.t.sol";
+import { D3MOperatorPlan } from "../../plans/D3MOperatorPlan.sol";
+
+contract D3MOperatorPlanTest is D3MPlanBaseTest {
+
+    bytes32 ilk = "DIRECT-PROTOCOL-A";
```
sunbreak1211, line 24: I don't think this is used anywhere.

sunbreak1211: Btw feel free to leave any non important comment on the tests for a later stage.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    function setUp() public {
+        baseInit("D3M4626TypePool");
+
+        vault = new ERC4626(ERC20(address(dai)), "dai vault", "DV");
+
+        setPoolContract(pool = new D3M4626TypePool(ILK, address(hub), address(dai), address(vault)));
+
+        pool.file("hub", address(hub));
+
```
telome, line 49:
```suggestion
```

src/tests/plans/D3MOperatorPlan.t.sol
hexonaut: removed

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,170 @@
...
+        vm.prank(address(hub)); pool.deposit(1);
+        
+        vat.cage();
+        vm.prank(address(hub)); pool.withdraw(1);
+
+        assertEq(pool.assetBalance(), 0);
+        assertEq(dai.balanceOf(address(hub)), 1);
+    }
+
+    function test_redeemable_returns_adai() public {
```
telome, line 104:
```suggestion
    function test_redeemable_returns_vault() public {
```

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: removed

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,170 @@
...
+        assertEq(pool.redeemable(), address(vault));
+    }
+
+    function test_exit_adai() public {
```
telome, line 106:
```suggestion
    function test_exit_vault_tokens() public {
```

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: fixed

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: fixed

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    function invariant_ilk_value() public {
+        assertEq(pool.ilk(), ILK);
+    }
+
+    function test_cannot_file_hub_no_auth() public {
+        pool.deny(address(this));
+        vm.expectRevert("D3M4626TypePool/not-authorized");
+        pool.file("hub", address(123));
+    }
```
sunbreak1211, line 71: This should be covered by `test_file_hub`.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,168 @@
...
+        deal(address(dai), address(pool), 1);
+        vm.prank(address(hub)); pool.deposit(1);
+        
+        vm.prank(address(hub)); pool.withdraw(1);
+        
+        assertEq(pool.assetBalance(), 0);
+        assertEq(dai.balanceOf(address(hub)), 1);
+    }
+
+    function test_withdraw_calls_vault_withdraw_vat_caged() public {
```
sunbreak1211, line 91: Not sure I understand the intention of this test.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,168 @@
...
+    function test_exit_vault_tokens() public {
+        deal(address(dai), address(this), 1e18);
+        vault.deposit(1e18, address(this));
+        uint256 tokens = vault.totalSupply();
+        vault.transfer(address(pool), tokens);
+        assertEq(vault.balanceOf(address(this)), 0);
+        assertEq(vault.balanceOf(address(pool)), tokens);
+
+        end.setArt(tokens);
+        vm.prank(address(hub)); pool.exit(address(this), tokens);
```
sunbreak1211, line 115: Semantically speaking we should be using the assets amount here, which is 1e18. As this is a super basic test they are the same amount. I think the compound test should be a better one to take as an example for this one.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,168 @@
...
+        assertEq(vault.balanceOf(address(this)), 0);
+        assertEq(vault.balanceOf(address(pool)), 1e18);
+
+        pool.quit(address(this));
+
+        assertEq(vault.balanceOf(address(this)), 1e18);
+        assertEq(vault.balanceOf(address(pool)), 0);
+    }
+
+    function test_assetBalance_gets_vault_balanceOf_pool() public {
```
sunbreak1211, line 134: These tests would be better if we could change the ratio of the total assets deposited again the vault supply.

sunbreak1211: I might be getting confused though, but seeing that we use `convertToAssets` for getting `assetBalance` makes me almost sure I'm right.

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -0,0 +1,168 @@
...
+        vault.transfer(address(pool), 1e18);
+        assertEq(dai.balanceOf(address(vault)), 1e18);
+        assertEq(vault.balanceOf(address(pool)), 1e18);
+
+        assertEq(pool.maxWithdraw(), 1e18);
+    }
+
+    function test_maxDeposit_returns_max_uint() public {
+        assertEq(pool.maxDeposit(), type(uint256).max);
+    }
```
sunbreak1211, line 167: I think we shouldn't be checking it returns max uint256 as that is a particularity of this vault deployment. We should be checking instead that it returns the same than the vault maxDeposit.
At least the test name shouldn't be tight to the max uint value.
Maybe we can just change the test name and add an extra check that the vault.maxDeposit is also max uint.

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+        require(msg.sender == operator, "D3MOperatorPlan/not-operator");
+
+        targetAssets = value;
+    }
+
+    function getTargetAssets(uint256) external override view returns (uint256) {
+        if (enabled == 0) return 0;
+
+        return targetAssets;
+    }
```
Rubilmax, line 86: Where is this value used within the D3M lifecycle?

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+        emit File(what, data);
+    }
+
+    function file(bytes32 what, uint256 data) external auth {
+        if (what == "enabled") {
+            require(data <= 1, "D3MOperatorPlan/invalid-value");
+            enabled = data;
+        } else revert("D3MOperatorPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
```
Rubilmax, line 74:
```suggestion
    function file(bytes32 what, bytes32 data) external auth {
        if (what == "operator") {
            operator = address(uint160(data));
        } else if (what == "enabled") {
            require(data <= 0x1, "D3MOperatorPlan/invalid-value");
            enabled = uint256(data);
        } else revert("D3MOperatorPlan/file-unrecognized-param");
        emit File(what, data);
    }
```

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+ *  @title D3M Operator Plan
+ *  @notice An operator sets the desired target assets.
+ */
+contract D3MOperatorPlan is ID3MPlan {
+
+    mapping (address => uint256) public wards;
+    uint256                      public enabled;
+
+    address public operator;
+    uint256 public targetAssets;
```
Rubilmax, line 31: Because it is public, this value can be read in place of `getTargetAssets` which would be interpreted incorrectly in case `enabled == 0` (because it seems it is expected to be interpreted as `0`, per `getTargetAssets`'s implementation)

src/tests/integration/MetaMorpho.t.sol
adhusson, line 1: If MM's supply queue has duplicate markets with some available supply then `maxDeposit` will strictly overestimate. So a wind limited by `maxDeposit` will revert.

This may break assumptions on `hub.exec` elsewhere.

src/plans/D3MOperatorPlan.sol
adhusson: https://github.com/makerdao/dss-direct-deposit/blob/master/src/D3MHub.sol#L295

Used by the hub to determine how much to wind/unwind.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,292 @@
...
+        return art;
+    }
+
+    function _divup(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        unchecked {
+            z = x != 0 ? ((x - 1) / y) + 1 : 0;
+        }
+    }
+
+    // function getInterestRateStrategy(address asset) internal view returns (address) {
```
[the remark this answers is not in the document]
MathisGD: everything commented is wip

src/plans/D3MOperatorPlan.sol
MathisGD: Actually, I think that it's better to enforce that the target assets is set back to zero before disabling the Plan. wdyt?

src/plans/D3MOperatorPlan.sol
MathisGD: Yes, but it's not the style elsewhere

https://github.com/makerdao/dss-direct-deposit/blob/5ed7835f76addc2aae1230ae575c8fee726a3bc7/src/plans/D3MAaveV2TypeRateTargetPlan.sol#L120-L129

src/plans/D3MOperatorPlan.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    function file(bytes32 what, uint256 data) external auth {
+        if (what == "enabled") {
+            require(data <= 1, "D3MOperatorPlan/invalid-value");
+            enabled = data;
+        } else revert("D3MOperatorPlan/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    function setTargetAssets(uint256 value) external {
```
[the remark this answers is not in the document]
MathisGD: image

src/plans/D3MOperatorPlan.sol
Rubilmax: I like this because it avoids side-effects which ultimately avoid unexpected/undesired behaviors
Is it compatible with the usual D3M life cycle though?

MathisGD: cc @hexonaut

src/tests/pools/D3M4626TypePool.t.sol
```diff
@@ -162,7 +157,11 @@ contract D3M4626TypePoolTest is D3MPoolBaseTest {
         assertEq(pool.maxWithdraw(), 1e18);
     }
 
-    function test_maxDeposit_returns_max_uint() public {
+    function invariant_pool_maxDeposit() public {
+        assertEq(pool.maxDeposit(), type(uint256).max);
+    }
+
+    function invariant_vault_maxDeposit() public {
         assertEq(pool.maxDeposit(), type(uint256).max);
```
sunbreak1211, line 165: Should be `vault` here.

src/plans/D3MOperatorPlan.sol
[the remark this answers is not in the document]
hexonaut: > so there's not automated assets targeting like Aave v2, computation is done offchain right?

yes this is correct

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: Why are these being resolved when the requested action is not complete? We let the issue author resolve the issues, so they can confirm the code author does the changes they want.

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: Re-opening so @sunbreak1211 can confirm this is complete.

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: Re-opening so @sunbreak1211 can confirm this is complete.

script/input/1/template-morpho.json
```diff
@@ -0,0 +1,14 @@
...
+    "poolType": "erc4626",
+    "planType": "operator",
+    "ilk": "DIRECT-SPARK-MORPHO-DAI",
+    "existingIlk": false,
+    "maxLine": 500000000,
+    "gap": 10000000,
+    "ttl": 21600,
+    "tau": 604800,
+    "vault": "0xB8C7F2a4B3bF76CC04bd55Ebc259b33a67b3b36d",
```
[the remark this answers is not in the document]
hexonaut: Again, why was this resolved @MathisGD ?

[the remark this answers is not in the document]
hexonaut: Also, due to the 2-step timelock we may actually do the configuration inside the DAO proxy.

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity ^0.8.14;
+pragma solidity 0.8.14;
```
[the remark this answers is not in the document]
hexonaut: Would also like to know this.

src/plans/D3MOperatorPlan.sol
hexonaut: Yes please no changes to the style.

src/plans/D3MOperatorPlan.sol
hexonaut: We need to be able to disable without requiring target assets is 0 in the case the multisig goes rogue.

src/tests/integration/MetaMorpho.t.sol
hexonaut: The SubDAO will need to ensure not to do this. I think it's fine.

script/input/1/template-morpho.json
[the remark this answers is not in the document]
hexonaut: Anyways, updated. Base case is we go with the new vault for now.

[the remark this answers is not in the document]
MathisGD: >Again, why was this resolved @MathisGD ?

my mistake, i resolved but didn't push

src/tests/integration/IntegrationBase.t.sol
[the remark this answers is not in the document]
MathisGD: I reverted it

src/tests/pools/D3M4626TypePool.t.sol
MathisGD: done here 3615bb4b48372a63557e87d6649bccf16dc09f9b

src/tests/integration/MetaMorpho.t.sol
adhusson: Worst case after a mistaken duplicate is a revert on exec, followed by a correcting setSupplyQueue (no timelock) then another exec. If that’s fine then 👍

src/plans/D3MOperatorPlan.sol
MathisGD: but any account that's capable of disabling the plan is also capable of setting themselves as operator to set the target assets at zero right?

anyway the current version is also fine to me, but it has the little tradeoff that @Rubilmax is mentioning

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -456,9 +456,9 @@ abstract contract IntegrationBaseTest is DssTest {
         // Rest of the liquidity can be withdrawn
         hub.exec(ilk);
         vow.heal(_min(vat.sin(address(vow)), vat.dai(address(vow))));
-        assertEq(vat.gem(ilk, address(end)), 0);
+        assertRoundingEq(vat.gem(ilk, address(end)), 0);
         assertEq(vat.sin(address(vow)), 0);
-        assertGe(vat.dai(address(vow)), prevDai); // As also probably accrues interest
+        //assertGe(vat.dai(address(vow)), prevDai); // As also probably accrues interest
```
MerlinEgalite, line 461: why is it commented?

src/tests/integration/IntegrationBase.t.sol
```diff
@@ -386,9 +386,9 @@ abstract contract IntegrationBaseTest is DssTest {
...
```
MerlinEgalite, line 391: same question as below?

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -14,14 +14,21 @@
...
 
-pragma solidity 0.8.14;
+pragma solidity ^0.8.14;
 
 import "forge-std/Test.sol";
 import "./IntegrationBase.t.sol";
 import "morpho-blue/src/interfaces/IMorpho.sol";
+import "forge-std/interfaces/IERC4626.sol";
+
+// Versioining issues with import, so it's inline
```
MerlinEgalite, line 24:
```suggestion
// Versioning issues with import, so it's inline
```

src/tests/integration/IntegrationBase.t.sol
MathisGD: failing with metamorpho for the moment

src/plans/D3MOperatorPlan.sol
Jean-Grimal: Or disabling the plan could also set back to zero target assets

src/tests/integration/IntegrationBase.t.sol
hexonaut: There is a rounding error making it off by one. These tests are overly strict imo.

src/tests/integration/IntegrationBase.t.sol
hexonaut: Same. Rounding error.

sunbreak1211:
@hexonaut I think is probably better to rollback Certora changes. So we keep this PR diff smaller. I'm trying to get a better version using a config file, and it seems we are still facing timeout issues anyway so we will have some CI failing.

src/deploy/D3MInit.sol
```diff
@@ -323,4 +362,13 @@ library D3MInit {
         plan.file("barb", compoundCfg.barb);
     }
 
+    function initOperatorPlan(
```
sunbreak1211, line 365: @hexonaut shouldn't we rely the `D3MMom` here?
I see it is not happening in the other plans, so not sure.

hexonaut: It's done in `initCommon` function.

hexonaut:
> @hexonaut I think is probably better to rollback Certora changes. So we keep this PR diff smaller. I'm trying to get a better version using a config file, and it seems we are still facing timeout issues anyway so we will have some CI failing.

Done

src/plans/D3MOperatorPlan.sol
hexonaut: > but any account that's capable of disabling the plan is also capable of setting themselves as operator to set the target assets at zero right?

Not in the way we have this constructed. This contract can trigger without a delay: https://github.com/makerdao/dss-direct-deposit/blob/master/src/D3MMom.sol#L76

sunbreak1211: I'd just make `targetAssets` a `private` variable.

src/tests/integration/IntegrationBase.t.sol
MerlinEgalite: alright thx

MerlinEgalite (approved):

Jean-Grimal (approved):

src/tests/integration/IntegrationBase.t.sol
sunbreak1211: We can replace the whole line by this one (removing the comment as I'm not sure why it mentions that probably accrues interest when the time is static):
```
assertApproxEqAbs(vat.dai(address(vow)), prevDai, roundingTolerance * RAY);
```

src/tests/integration/IntegrationBase.t.sol
sunbreak1211: We can replace the whole line by this one (removing the comment as I'm not sure why it mentions that probably accrues interest when the time is static):
```
assertApproxEqAbs(vat.dai(address(vow)), prevDai, roundingTolerance * RAY);
```

src/tests/integration/IntegrationBase.t.sol
sunbreak1211: Same here, replace it by:
```
assertApproxEqAbs(vat.dai(address(vow)), prevDai, roundingTolerance * RAY);
```

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,210 @@
...
+    address operator = makeAddr("operator");
+    uint256 constant startingAmount = 5_000_000 * WAD;
+
+    function setUp() public {
+        baseInit();
+
+        vm.createSelectFork(vm.envString("ETH_RPC_URL"), 19456934);
+
+        // Need to increase to 3
+        roundingTolerance = 3;
```
sunbreak1211, line 75: All the tests are passing for me setting this to `1`.

hexonaut: Must have been resolved with the other changes. Reverted.

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: I think this is resolved? @sunbreak1211

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: I think this is resolved. @sunbreak1211

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: Also think this is resolved. @sunbreak1211

src/plans/D3MOperatorPlan.sol
hexonaut: Going to leave this for now unless you guys feel strongly. Code is in audit, and this is somewhat of a superficial change on a backend contract which can be upgraded later if desired.

src/tests/pools/D3M4626TypePool.t.sol
hexonaut: Can be resolved I think. @sunbreak1211

Rubilmax (approved):

src/tests/integration/IntegrationBase.t.sol
hexonaut: re-enabled, but relaxed the assert

src/tests/integration/IntegrationBase.t.sol
hexonaut: re-enabled, but relaxed the assert

.github/workflows/certora.yml
```diff
@@ -26,11 +26,11 @@ jobs:
...
       - name: Checkout
         uses: actions/checkout@v3
-      
+
       - uses: actions/setup-java@v3
         with:
-          distribution: 'zulu'
-          java-version: '11'
+          distribution: "zulu"
+          java-version: "11"
```
sunbreak1211, line 33: We might want to roll back this change as well as a bit weird to have it here and actually doesn't change anything on how it works.

src/tests/integration/IntegrationBase.t.sol
sunbreak1211: LGTM

src/tests/integration/IntegrationBase.t.sol
sunbreak1211: LGTM

hexonaut:
Merged Certora CI fixes.

.github/workflows/certora.yml
QGarchery: Yes it's not related to this PR, it was introduced automatically by the formatter (prettier in this case)

.github/workflows/certora.yml
```diff
@@ -26,11 +26,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      
+
```
MathisGD, line 29:
```suggestion
      
```

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        (,,, uint256 line,) = VatAbstract(vat).ilks(ilk);
+        vm.assume(target1 < line * 1e18 / 1e45);
+        
+        adjustDebt(int256(target1));
+        
+        vm.assume(target2 <= pool.assetBalance());
+        
+        uint256 marketSupplyMiddle = morpho.market(marketParams.id()).totalSupplyAssets;
+        
+        uint256 depositedAssets1 = min(target1, line * 1e18 / 1e45);
```
sunbreak1211, line 245: There is a `vm.assume` that `target1` is less than the `line`, so this will be target1.
I'd remove `depositedAssests1`, just use `target1` and rename `depositedAssets2` to `depositedAssets`.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        uint256 depositedAssets = min(d3mDeposit, min(line * 1e18 / 1e45, cap));
+        
+        assertEq(morpho.market(marketParams.id()).totalSupplyAssets, marketSupplyBefore + depositedAssets);
+    }
+
+    function testWithdrawLiquid(uint256 target1, uint256 target2) public {
+        uint256 marketSupplyStart = morpho.market(marketParams.id()).totalSupplyAssets;
+        
+        (,,, uint256 line,) = VatAbstract(vat).ilks(ilk);
+        vm.assume(target1 < line * 1e18 / 1e45);
```
hexonaut, line 237: Replace these with `_bound()`. `vm.assume()` is too slow with lots of fuzz runs.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        
+        adjustDebt(int256(target1));
+        
+        vm.assume(target2 <= pool.assetBalance());
```
hexonaut, line 241: Replace these with `_bound()`. `vm.assume()` is too slow with lots of fuzz runs.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        assertGe(marketSupplyMiddle, marketSupplyStart);
+        assertLe(marketSupplyEnd, marketSupplyMiddle);
+        assertEq(marketSupplyEnd, marketSupplyMiddle - expectedWithdraw);
+        assertEq(marketSupplyEnd, marketSupplyStart + depositedAssets2);
+    }
+
+    function testWithdrawIlliquid(uint256 target1, uint256 target2, uint256 borrow) public {
+        uint256 marketSupplyStart = morpho.market(marketParams.id()).totalSupplyAssets;
+        uint256 marketBorrowStart = morpho.market(marketParams.id()).totalBorrowAssets;
+        vm.assume(borrow < marketSupplyStart - marketBorrowStart);
```
hexonaut, line 265: Replace these with `_bound()`. `vm.assume()` is too slow with lots of fuzz runs.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        
+        (,,, uint256 line,) = VatAbstract(vat).ilks(ilk);
+        vm.assume(target1 < line * 1e18 / 1e45);
```
hexonaut, line 268: Replace these with `_bound()`. `vm.assume()` is too slow with lots of fuzz runs.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        
+        adjustDebt(int256(target1));
+        adjustLiquidity(-int256(borrow));
+        
+        vm.assume(target2 <= pool.assetBalance());
```
hexonaut, line 273: Replace these with `_bound()`. `vm.assume()` is too slow with lots of fuzz runs.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+    D3MOperatorPlan plan;
+    D3M4626TypePool pool;
+
+    // sUSDe/USDC (91.5%).
+    MarketParams public marketParams = MarketParams({
+        loanToken: 0x6B175474E89094C44Da98b954EedeAC495271d0F,
+        collateralToken: 0x9D39A5DE30e57443BfF2A8307A4256c8797A3497,
+        oracle: 0x5D916980D5Ae1737a8330Bf24dF812b2911Aae25,
+        irm: 0x870aC11D48B15DB9a138Cf899d20F13F79Ba00BC,
+        lltv: 915000000000000000
```
QGarchery, line 61: We can probably have less "magic" values, for example reusing the `sUsde` variable above

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        assertEq(marketSupplyEnd, marketSupplyStart + depositedAssets2);
+    }
+}
+
+function min(uint256 x, uint256 y) pure returns (uint256) {
+    return x < y ? x : y;
+}
+function max(uint256 x, uint256 y) pure returns (uint256) {
+    return x > y ? x : y;
+}
```
QGarchery, line 300: MetaMorpho imports `Math` (from OZ, through ERC4626), we could reuse that

QGarchery (approved):

src/tests/integration/MetaMorpho.t.sol
hexonaut: We are actually going to remove the dependencies before this gets merged, so best to leave this.

src/tests/integration/MetaMorpho.t.sol
MathisGD: done for all vm.assume

src/tests/integration/MetaMorpho.t.sol
MathisGD: fixed

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,300 @@
...
+        vm.assume(target1 < line * 1e18 / 1e45);
+        
+        adjustDebt(int256(target1));
+        adjustLiquidity(-int256(borrow));
+        
+        vm.assume(target2 <= pool.assetBalance());
+        
+        uint256 marketSupplyMiddle = morpho.market(marketParams.id()).totalSupplyAssets;
+        
+        uint256 depositedAssets1 = min(target1, line * 1e18 / 1e45);
```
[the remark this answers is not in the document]
MathisGD: fixed in 0ddf50f989978d79ec33b62fac9016a1a1b5cd74

src/tests/integration/MetaMorpho.t.sol
MathisGD: done in 22c79f68833bbc5d6cfd0e99cf7060bd41a7a5a6

hexonaut (commented):
New tests look good. Fixed previous issues.

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,479 @@
...
+        assertEq(plan.targetAssets(), d3mDeposit);
+
+        assertEq(morpho.market(marketParams.id()).totalSupplyAssets, marketSupplyBefore + maxLineScaled);
+
+        assertEq(spDai.balanceOf(address(pool)), maxLineScaled);
+        assertEq(spDai.totalSupply(), spDaiTotalSupplyBefore + maxLineScaled);
+        assertEq(
+            spDai.totalAssets(), spDaiTotalAssetsBefore + morpho.expectedSupplyAssets(marketParams, address(spDai))
+        );
+        assertEq(spDai.maxDeposit(address(pool)), spDaiMaxDepositBefore - maxLineScaled);
```
sunbreak1211, line 291: Any reason to not have added the same check than the prev two:
```
assertEq(pool.assetBalance(), UtilsLib.zeroFloorSub(d3mDeposit, 1));
```

src/tests/integration/MetaMorpho.t.sol
```diff
@@ -0,0 +1,479 @@
...
+        plan.setTargetAssets(target2);
+        hub.exec(ilk);
+
+        uint256 marketSupplyEnd = morpho.market(marketParams.id()).totalSupplyAssets;
+
+        assertEq(marketSupplyEnd, marketSupplyMiddle - expectedWithdraw);
+        assertEq(marketSupplyEnd, marketSupplyStart + depositedAssets2);
+    }
+
+    function testWithdrawIlliquid(uint256 target1, uint256 target2, uint256 borrow) public {
```
sunbreak1211, line 448: I'm a bit confused about this test. I'd assume that if we are testing withdraw for an illiquid market is because we need to withdraw more than the pool.maxWithdraw();
Then the following assert should pass always:
```
assertGt(pool.assetBalance() - target2, pool.maxWithdraw());
```
(which it doesn't)
We could even be executing a full run without any illiquid case.

I think we should start bounding the borrow value to be minimum the amount that was already available (marketSupplyStart - marketBorrowStart) and then make sure poolAsset() - target 2 > available funds.

Although I might be understating wrongly the meaning of this test.

sunbreak1211:
@hexonaut let's add the date to the audit file name.

src/tests/integration/MetaMorpho.t.sol
MathisGD: fixed in 576c75cb8a76dbbd3390335920307bd3f09b7647

src/tests/integration/MetaMorpho.t.sol
MathisGD: you are right, fixed in 576c75cb8a76dbbd3390335920307bd3f09b7647

now we try to withdraw more than the market's liquidity

```solidity
target2 = bound(target2, 0, pool.assetBalance() - liquidityMiddle);
```

MathisGD:
See #121 for the failing tests

SidestreamColdMelon:
While writing integration test for this module in the [2024-03-26 spell](https://github.com/makerdao/spells-mainnet/pull/402), I noticed that for the previous D3Ms we've tested disabling the plan with via MOM and effectively bringing `art` and `ink` to ~0, see ([SPARK](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2023-04-28-DssSpell/DssSpell.t.sol#L1203-L1206), [AAVEV2](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2023-02-22-DssSpell/DssSpell.t.sol#L739-L742), [COMPV2](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2022-12-02-DssSpell/DssSpell.t.sol#L792-L795)). Doing the same with this module [currently doesn't seem to work](https://github.com/makerdao/spells-mainnet/pull/402/commits/28809ffa21a3aa06c7217105d717d7a02fc6c0af). Am I missing something obvious?

hexonaut:
> While writing integration test for this module in the [2024-03-26 spell](https://github.com/makerdao/spells-mainnet/pull/402), I noticed that for the previous D3Ms we've tested disabling the plan with via MOM and effectively bringing `art` and `ink` to ~0, see ([SPARK](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2023-04-28-DssSpell/DssSpell.t.sol#L1203-L1206), [AAVEV2](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2023-02-22-DssSpell/DssSpell.t.sol#L739-L742), [COMPV2](https://github.com/makerdao/spells-mainnet/blob/a4e7963172c6d3691bacb5e487a093e69b16c4fd/archive/2022-12-02-DssSpell/DssSpell.t.sol#L792-L795)). Doing the same with this module [currently doesn't seem to work](https://github.com/makerdao/spells-mainnet/pull/402/commits/28809ffa21a3aa06c7217105d717d7a02fc6c0af). Am I missing something obvious?

Sorry forgot to mention about this. I added a response to fix: https://github.com/makerdao/spells-mainnet/pull/402#pullrequestreview-1955933590

MerlinEgalite (approved):

SidestreamColdMelon:
FYI: `D3MInit.sol` [imports DssInstance](https://github.com/makerdao/dss-direct-deposit/blob/8a18cb35c19e30f9527f7b1489f79cfcf33a7fa5/src/deploy/D3MInit.sol#L23) from the [dss-test](https://github.com/makerdao/dss-test/) and requires it to be passed to the init scripts. This implies importing `MCD` library in the spell, which in turn imports `GodMode` and `vm` contracts, which eventually leads to them [being deployed with the spell](https://etherscan.io/address/0xcD672aCc9885796a19b4bAf03Dba46c8cdB0882B#code#L4086). This at least significantly increases deployment transaction cost. Note that this pattern is present across past spells as well, see [spark D3M being deployed in 2023-04-28](https://etherscan.io/address/0x77107F74bf30250aFFada0fbD09fa517658B4916#code#L3639).

sunbreak1211:
@hexonaut two comments:
- very minor: can we change the audit file name to: `2024_03_Spearbit-report-maker-d3m-implementation-for-morpho-review`. It seems easier for who is looking at the list to identify the file.
- who and when will be taking care to remove the dependencies so we can finish the PR and merge it?

hexonaut:
> @hexonaut two comments:
> 
> * very minor: can we change the audit file name to: `2024_03_Spearbit-report-maker-d3m-implementation-for-morpho-review`. It seems easier for who is looking at the list to identify the file.
> * who and when will be taking care to remove the dependencies so we can finish the PR and merge it?

Renamed audit file.

@MathisGD are you able to copy in any dependencies to remove the solmate and metamorpho libraries?

hexonaut (commented):
Morpho dep has been removed/inlined. This is good to merge now imo.

sunbreak1211 (approved):
LGTM

MerlinEgalite (approved):

hexonaut:
Going to merge this. Additional cleanup can be done via separate PRs.

